PR #1, opened 2019-08-22
Updated dream.py to use a function for generating pixel positions

SimonHeybrock:


dream.py
```diff
@@ -20,16 +20,73 @@ def make_component_info(source_pos, sample_pos):
...
+    """
+    
+    if not n_pixel % n_rows == 0:
+        print("The number of pixels can not be evenly distributed into the nunmber of rows")
+    
+    n_pixel_per_row = int(n_pixel/n_rows)
+    if n_rows == 1:
+        heights = [center[1]]
+    else:
+        heights = np.linspace(-0.5*height, 0.5*height, n_rows) + center[1]
```
SimonHeybrock: This appears to assume that the primary beam is aligned with the axis (and thus the cylinder is aligned around that? Maybe we should drop the assumption, or verify it and fail if it is not fulfilled? Otherwise we may run into hard to debug surprises later on.

dream.py
```diff
@@ -20,16 +20,73 @@ def make_component_info(source_pos, sample_pos):
...
+    # Create a numpy array that describes detector positions
+    pixel_positions = np.zeros((n_pixel_per_row*n_rows, 3))
+    ttheta_array = np.linspace(ttheta_min, ttheta_max, n_pixel_per_row)
+
+    for row in range(n_rows):
+        from_index = row*n_pixel_per_row
+        to_index = (row+1)*n_pixel_per_row
+        pixel_positions[from_index:to_index,0] = radius*np.sin(3.14159/180*ttheta_array) + center[0]
+        pixel_positions[from_index:to_index,1] = heights[row]
+        pixel_positions[from_index:to_index,2] = radius*np.cos(3.14159/180*ttheta_array) + center[2]
```
SimonHeybrock: `np.pi`?

dream.py
```diff
@@ -46,23 +103,22 @@ def make_component_info(source_pos, sample_pos):
...
+#print(d)
 
 dspacing = sc.neutron.convert(d, Dim.Tof, Dim.DSpacing)
 print(dspacing)
 
 # Converting event data to histogram
-hist = sc.histogram(dspacing, dspacing.coords[Dim.DSpacing])
-print(hist)
+#hist = sc.histogram(dspacing, dspacing.coords[Dim.DSpacing])
+#print(hist)
```
SimonHeybrock: Is this not working any more with your changes, or why is it commented now?

dream.py
mads-bertelsen: Yes, this was built with the idea of a classic reactor detector bank. The real DREAM detector bank is rotated in another way. Would you like it to be adjustable?

dream.py
mads-bertelsen: Wasn't aware that np included constants, neat.

dream.py
mads-bertelsen: scipp complains about the dspacing bin edges not being sorted to make a histogram. I would look more into this once I have generated some data to be plotted.

mads-bertelsen: scipp complains about the dspacing bin edges not being sorted to make a histogram. I would look more into this once I have generated some data to be plotted.

mads-bertelsen:
The code was updated according to the requests, and is now structured as a class instead. I think this will both limits the ways the user can misuse it and provide a better base for the future work in generating events and data reduction.

SimonHeybrock (commented):
:+1: for encapsulating the instrument generation in a class.

I am not so sure about wrapping the event-generation, conversion, and histogramming. It does not seem to add anything, apart from making it unclear what happens?
- You will end up having two references to a dataset, so it will not be easy to "drop" a dataset, freeing up the memory.
- You are hiding what is actually happening, removing a lot of the flexibility.
- The user has two memorize two APIs, that of scipp and that of your class.

We can change this as part of future changes though.

PR #2, opened 2019-08-26
Added generation of random powder data.

mads-bertelsen:
A number of random Bragg peaks are defined with a given range of
d spacings, delta d over d values and relative strengths.
The wavelength frame is defined and the Bragg peaks available for
each pixel is identified. A poisson distributed random number of
events are then drawn from each bragg peak, each with a gaussion
distribution of d spacings. The instrument length, detector
distance and a random source origin time is used to calculate the
corresponding time for the event.

This provides a dataset with somewhat realistic correlations
between time and theta value of each pixel. There is no background,
but this could be added if desired.

The performance scaling of the code is problematic for the number
of pixels generated, but the scaling for number of events is good.

1E7 events in 1E5 pixels takes 15 seconds on my laptop. Going to
5E5 pixels the run time is around 6 minutes.

### DIFF
--- a/dream.py
+++ b/dream.py
@@ -194,25 +194,6 @@ class DreamTest:
 
         return pixel_coords
 
-    def generate_data_test(self, n_events):
-        """
-        This method generates n_events of sparse tof data.
-        To be expanded to generate plausible data into some Bragg peaks in the future
-        
-        Parameters
-        ----------
-            n_events: int
-                Number of events to generate and add to current dataset self.d
-        """
-
-        # Add some events
-        # Note: d.coords[Dim.Tof] gives the "dense" TOF coord, not the event-TOFs
-        self.d['sample'].coords[Dim.Tof][Dim.Position, 0].values = np.arange(n_events)
-        self.d['sample'].coords[Dim.Tof][Dim.Position, 1].values = np.arange(n_events)
-        #self.d['sample'].coords[Dim.Tof][Dim.Position, [1,2]].values = np.array((np.arange(n_events), np.array([3, 3, 3])))
-        # The following should be equivalent but does not work yet, see scipp/#290
-        # d['sample'].coords[Dim.Tof].values[1] = np.arange(10)
-         
     def _generate_random_sample(self, n_peaks, d_range=[0.5, 10], deltad_over_d_range=[0.1, 0.5], strength_range=[1, 10]):
         """
         Generate a random series of Bragg peaks. They will be
--- a/dream.py
+++ b/dream.py
@@ -1,12 +1,66 @@
+import time
 import numpy as np
 import scipp as sc
 from scipp import Dim
+
+class bragg_peak:
+    """
+    Class to hold a Bragg peak generator with a given d_spacing,
+    width and strength. The strength is a unphysical description
+    of its strength, and has no absolute meaning, but is relative
+    to the total strength observed.
+    The Bragg peak can generate a numpy array of d_spacing values
+    with a length found from the strength of the peak and a given
+    conversion factor between strength and expected events.
+    The number of events is drawn from a poisson distribution.
+    """
+    def __init__(self, d_spacing, d_width, strength):
+        """
+        Initialize a Bragg peak
+        
+        Parameters
+        ----------
+            d_spacing: float
+                Central d_spacing of the peak
+            
+            d_width: float
+                Width of the gaussian d_spacing distribution
+            
+            strength: float
+                Strength of the Bragg peak (relative)
+        """
+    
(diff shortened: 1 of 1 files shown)


dream.py
```diff
@@ -130,7 +194,7 @@ def _make_cylinder_coords(self, n_pixel, n_rows=1, height=0.1, radius=4,
 
         return pixel_coords
 
-    def generate_data(self, n_events):
+    def generate_data_test(self, n_events):
```
SimonHeybrock: Is this function still useful? It was meant to demo the API for adding sparse data, but can probably be removed now that you have a proper generator?

dream.py
```diff
@@ -144,40 +208,193 @@ def generate_data(self, n_events):
...
+                Number of events to generate
+        
+            wavelength_width: float [Å]
+                Width of wavelength frame of the instrument, 3.6 Å should fill the frame with perfect source
+            
+            wavelength_center: float [Å]
+                Center of wavelength frame of the instrument, typicall around 2.5-3.0 Å
+            
+            effective_pulse_length: float [s]
+                Effective pulse length (after choppers have modified the pulse), between 10 us and 2.86 ms
```
SimonHeybrock, line 256: More of an idea than a request: `scipp` supports scalar variables with units. Would it make sense to use them here for the input parameters? This would add extra safety (and reduce the need for documentation?). So instead of giving, e.g., `wavelength_center=3.0` we could use `wavelength_center=sc.Variable(value=3.0, unit=sc.units.angstrom)`.

If this is useful, I am also considering addition of operators to make this more convenient: `wavelength_center=3.0 * sc.units.angstrom` could in principle be supported, I think (avoiding the need to write `sc.Variable(...)`).

dream.py
```diff
@@ -144,40 +208,193 @@ def generate_data(self, n_events):
...
+            
+        # Now we have pixel_strength as indicator for relative intensity in each peak (assuming relatively narrow peaks)
+        total_strength = np.sum(pixel_strengths)
+        events_per_strength = n_events / total_strength
 
-    #def data_reduction(self):
-        # "DiffractionFocussing" == sum? (not available yet)
-        # focussed = sc.sum(hist, Dim.Position)
+        # Calculate distance from source to each pixel when scattered in sample
+        source_to_pixel = self.source_sample_dist + position_lengths
```
SimonHeybrock, line 294: We have exactly this code (computing scattering angles, L2, ...) in the implementation of the unit conversions in C++. We should discuss the best way of making this available in Python. Either as part of a task in your next sprint, or independently.

dream.py
```diff
@@ -144,40 +208,193 @@ def generate_data(self, n_events):
...
+        for pixel_id in range(n_positions):
+        
+            if progress_bar and pixel_id % progress_fraction == 0:
+                print(str(progress_count*10) + "%")
+                progress_count += 1
+            
+            allowed_bragg_peaks = allowed[pixel_id,:]
+            wavelengths = []
+            for bragg_peak_id in np.arange(len(allowed_bragg_peaks))[allowed_bragg_peaks]:
+                returned_wavelengths = 2.0*np.sin(thetas[pixel_id])*self.sample_bragg_peaks[bragg_peak_id].generate_events(events_per_strength)
```
SimonHeybrock, line 312: Is the call to `generate_events` the slow part? Would it help to just use the same random numbers all the name, and simply take a slice of required length, and then scale/offset it?

SimonHeybrock (commented):
This looks good, :+1: for documenting the functions. I have mainly some questions / things to discuss:

dream.py
mads-bertelsen: It can be removed, will update the branch.

dream.py
mads-bertelsen: I believe the root of the performance problem is that it is two nested for loops that need to work with small arrays. Due to the input interface for sparse data, a for loop over detector pixels is mandatory. I think the problem is to hold the entire pixel array in memory, as the execution time scales almost only with this parameter and changes slope at a critical length. See quick graph of execution times.
As I understand it, your proposed fix would improve the scaling with n_events, but the problem lies with the scaling of n_pixels.

IMG_0527

dream.py
mads-bertelsen: This does sound like a nice idea and something I will try to use when working with scipp in the future. I don't think it is necessary in a script for internal testing though?

SimonHeybrock: Yes, not necessary, but it may actually be easier, in particular when combined with operating on the beamline geometry helpers, which return `scipp.Variable` objects.

But as I said, there is not need to include this now, unless you want to.

PR #3, opened 2019-09-23
Function that can load a nexus event file using scipp/mantid and

SimonHeybrock:
expand this to more pixels / events by replicating the existing
pixels / events. Random noise is added to the event times in order
to avoid perfect overlap in that regard.

A quick test script is available that just runs the function with
a reasonable set of inputs.

No data file is included as all data files we have available can
not be shared due to the ISIS data policy. A datafile for inhouse
testing is available through ownCloud.

### DIFF
--- a/expand_load.py
+++ b/expand_load.py
@@ -39,7 +39,8 @@ def expand_data_file(filename, n_pixels, n_events=None, time_noise_us=200, verbo
     """
     # Make sure n_pixels and n_events are ints or can at least be converted
     n_pixels = int(n_pixels)
-    n_events = int(n_events)
+    if n_events is not None:
+        n_events = int(n_events)
 
     d = sc.Dataset()
     d["loaded_data"] = load(filename=filename, load_pulse_times=False)
@@ -74,7 +75,12 @@ def expand_data_file(filename, n_pixels, n_events=None, time_noise_us=200, verbo
     if remaining_length > 0:
         combined_var = sc.concatenate(combined_var, positions_var[Dim.Position, 0:remaining_length], Dim.Position)
 
+    original_Tof = d["loaded_data"].coords[Dim.Tof]
+
+    # Create new dataset with new positions
     ds = sc.Dataset(coords={Dim.Position: combined_var})
+    # Keep component_info from loaded data manually
+    ds.labels["component_info"] = d.labels["component_info"]
 
     if verbose:
         print("Generated dataset before tof events are added")
@@ -89,11 +95,12 @@ def expand_data_file(filename, n_pixels, n_events=None, time_noise_us=200, verbo
         original_id = 0
         for pixel_id in range(n_pixels):
             # For each pixel, grab the events from the original file
-            pos_n_events = len(np.array(d["loaded_data"].coords[Dim.Tof].values[original_id]))
+            #pos_n_events = len(np.array(d["loaded_data"].coords[Dim.Tof].values[original_id]))
+            pos_n_events = len(d["loaded_data"].coords[Dim.Tof].values[original_id])
 
             # If any events present, add random noise and append them as sparse data
             if pos_n_events > 0:
-                noise = np.random.normal(pos_n_events)*time_noise_us
+                noise = np.random.normal(pos_n_events, scale=time_noise_us)
                 new_values = np.array(d["loaded_data"].coords[Dim.Tof].values[original_id]) + noise
 
                 ds["measurement"].coords[Dim.Tof][Dim.Position, pixel_id].values.extend(new_values)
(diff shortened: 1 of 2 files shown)


expand_load.py
```diff
@@ -0,0 +1,135 @@
...
+    # Adding sparse data to dataset
+    tofs = sc.Variable(dims=[Dim.Position, Dim.Tof], shape=[n_pixels, sc.Dimensions.Sparse], unit=sc.units.us)
+    ds["measurement"] = sc.DataArray(coords={Dim.Tof: tofs})
+
+    added_events = 0
+    while n_events is None or added_events < n_events: # keep adding sparse data until enough events reached
+        original_id = 0
+        for pixel_id in range(n_pixels):
+            # For each pixel, grab the events from the original file
+            pos_n_events = len(np.array(d["loaded_data"].coords[Dim.Tof].values[original_id]))
```
SimonHeybrock: I recommend working with `DataArray` instead of `Dataset`. Then you can avoid the artificial (and more lengthy) explicit item names such as `"loaded_data"`.

Why did you convert to `np.array` before taking `len`?

test_expand_load.py
```diff
@@ -0,0 +1,18 @@
...
+conda create -n scipp_mantid_env -c scipp/label/dev -c mantid scipp mantid-framework python=3.6
+"""
+
+data = expand_load.expand_data_file('../../wish_event_data/WISH00043525.nxs', 3E5, n_events=6E8, verbose=True)
+
+# tested cases
+#data = expand_load.expand_data_file('../../wish_event_data/WISH00043525.nxs', 6E5)
+#data = expand_load.expand_data_file('../../wish_event_data/WISH00043525.nxs', 3E5, n_events=6E8, verbose=True)
+#data = expand_load.expand_data_file('../../wish_event_data/WISH00043525.nxs', 3E5, n_events=6E8, time_noise_us=500, verbose=True)
+#data = expand_load.expand_data_file('../../wish_event_data/WISH00043525.nxs', 3E5, n_events=6E8, time_noise_us=100, verbose=True)
```
SimonHeybrock: I suggest we remove the hardcoded filenames and use a command-line argument as the filename to load, so we can include this in `master`.

expand_load.py
```diff
@@ -0,0 +1,135 @@
...
+
+            # If any events present, add random noise and append them as sparse data
+            if pos_n_events > 0:
+                noise = np.random.normal(pos_n_events)*time_noise_us
```
SimonHeybrock: Was there a reason for not using the numpy parameter `scale=time_noise_us`?

SimonHeybrock (commented):
@mads-bertelsen Some small comments below.

Additionally, the main thing I think we need to fix is that we lose extra information from the loaded input that may be essential. For example the `component_info` labels are lost, which contain source and sample positions. That is, we could not convert units of the created dataset.

To fix this, I think we should try to be future-proof (so you do not need to come back and "fix" this function 5 times): In the coming weeks/months we will probably extend the existing converters and add other information that is present in the Mantid workspace. The main example that comes to mind is information about the sample and other beamline logs. That is, we need to avoid implementing `expand_data_file` in a "white list" approach (copying things explicitly by hand). I think one way to do this is to just create a copy of the original `Dataset` (or `DataArray`), `del` the things that we know depend on the spectrum (TOFs, coords[`Dim.Position`], and labels[`spectrum_number`]), and add them back in (as you do now).

expand_load.py
mads-bertelsen: No reason, just didn't know that option existed. Will use that.

test_expand_load.py
mads-bertelsen: It has been updated with a command line interface including all options and documentation with both doctoring and -h option.

expand_load.py
mads-bertelsen: Will try to use DataArray in the future, but I don't see how I could have done it in that particular instance? Should I have saved  d["loaded_data"] as another variable and worked with that?

I didn't know scipp had len functionality built in, it is never mentioned in the documentation.

mads-bertelsen:
All concerns have been addressed, except the change to a "blacklist" approach over the current "whitelist" approach, which was not possible due to scipp not being able to delete coords / labels. The loaded component_info was however forwarded to the new dataset.

SimonHeybrock:
I have created https://github.com/scipp/scipp/issues/564 for the missing functionality.

expand_load.py
SimonHeybrock: Something like:
```python
loaded_data = sc.neutron.load(...)
output = sc.DataArray(coords=...)
```

mads-bertelsen: I see, thought you meant internally, but the returned result can just be a DataArray. This branch has been updated with this improvement now. Ready to merge?

PR #4, opened 2019-10-04
Bug fix, degrees used instead of radians in trig functions

mads-bertelsen:
Fixed two occurrences of trigonometric functions having degrees used instead of radians. The results have been plotted and now look as expected.

Some auto formatting seems to have happened in my editor, only two lines are actually changed.


SimonHeybrock:
@mads-bertelsen Should this be merged? I see this only now.

mads-bertelsen:
@SimonHeybrock Yes, it is just a small bugfix.

PR #5, opened 2019-10-24
Removed test code from module and seperated dspacing conversion

mads-bertelsen:
Removed test code from dream.py and moved conversion to dspacing into a separate module.


mads-bertelsen:
@SimonHeybrock Forgot to tag you on this, probably good practice!

SimonHeybrock:
@mads-bertelsen Thanks. Note that you can request reviewers in github (top right).

PR #7, opened 2019-10-28
Das 137

mads-bertelsen:
Upload of a Jupyter Notebook that demos the capabilities of the dream module. It was uploaded on request of Simon Heybrock. The code relies on recent changes to the dream module, so the other active pull requests should be considered first.

The notebook was cleared before uploading as instructed.

### DIFF
--- a/dream_demo.ipynb
+++ b/dream_demo.ipynb
@@ -2,7 +2,7 @@
  "cells": [
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": null,
    "metadata": {},
    "outputs": [],
    "source": [
@@ -32,38 +32,9 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 6,
+   "execution_count": null,
    "metadata": {},
-   "outputs": [
-    {
-     "name": "stdout",
-     "output_type": "stream",
-     "text": [
-      "Dataset with pixels initialized\n",
-      "<scipp.Dataset>\n",
-      "Dimensions: {{Dim.Tof, 10}, {Dim.Position, 20000}}\n",
-      "Coordinates:\n",
-      "    Dim.Position              vector_3_double  [m]              (Dim.Position)  [(0.323524, 0.000000, 1.207407), (0.323677, 0.000000, 1.207366), ..., (0.427674, 0.000000, -1.174562), (0.427525, 0.000000, -1.174616)]\n",
-      "    Dim.Tof                   double     [μs]            (Dim.Tof)  [0.000000, 1.000000, ..., 8.000000, 9.000000]\n",
-      "Labels:\n",
-      "    component_info            Dataset    [dimensionless]  ()  [Dataset]\n",
-      "Data:\n",
-      "    sample\n",
-      "        Sparse coordinate:\n",
-      "            Dim.Tof           double     [μs]            (Dim.Position, Dim.Tof [sparse])  [[], [], ..., [], []]\n",
-      "\n",
-      "\n",
-      "Calibration dataset:\n",
-      "<scipp.Dataset>\n",
-      "Dimensions: {{Dim.Position, 20000}}\n",
-      "Data:\n",
-      "    difc                      double     [Å^-1 μs]     (Dim.Position)  [5133.898105, 5136.365414, ..., 38734.342550, 38734.774770]\n",
-      "    tzero                     double     [μs]            (Dim.Position)  [0.000000, 0.000000, ..., 0.000000, 0.000000]\n",
-      "\n",
-      "\n"
-     ]
-    }
-   ],
+   "outputs": [],
    "source": [
     "# Script for demonstration of event generation\n",
     "n_pixel = int(2E4)\n",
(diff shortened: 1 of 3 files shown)


SimonHeybrock:
> The notebook was cleared before uploading as instructed.

@mads-bertelsen The notebook in this branch *does* contain outputs. Can you double check that you clear them before saving?

mads-bertelsen:
@SimonHeybrock I don't understand how that happened. I see it wasn't cleared, even in my local version, but clearing it again in exactly the same way worked. Sorry about that, good catch!

dream_demo.ipynb
```diff
@@ -0,0 +1,261 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Script for demonstration of event generation\n",
+    "n_pixel = int(2E4)\n",
+    "n_events = int(2E6)\n",
+    "\n",
+    "Dream = dream.DreamTest(n_pixel) # Create a fake dream powder diffractometer with detector positions and calibration data\n",
```
SimonHeybrock: Uppercase names are usually only used for classes, not for variables (see https://www.python.org/dev/peps/pep-0008/). Reading `Dream`, most people would think it is a class, but it is not.

dream_demo.ipynb
```diff
@@ -0,0 +1,261 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "dspacing = powder_reduction.powder_reduction(Dream.data, calibration=Dream.cal)\n",
+    "print(dspacing) # Show the converted dataset"
```
SimonHeybrock: Do you feel that comments like this are necessary in a notebook? Isn't it self documenting, given that the output will appear immediately below?

dream_demo.ipynb
```diff
@@ -0,0 +1,261 @@
...
+   "source": [
+    "print(dspacing.dims) # Show the dimensionality of the dspacing dataset"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Plotting the dspacing dataset\n",
+    "First the raw data is collected from each of the sparse channels in the detector array. Next `numpy` is used to histogram the data, after which the data is plotted with `matplotlib`. We see a diffraction pattern with peaks in the specified interval."
```
SimonHeybrock: Why are you using `numpy` for histogramming instead of `scipp`?

dream_demo.ipynb
```diff
@@ -0,0 +1,261 @@
...
+   "outputs": [],
+   "source": [
+    "# Plot using scipp feature\n",
+    "import numpy as np\n",
+    "from scipp.plot import plot\n",
+    "%matplotlib inline\n",
+    "\n",
+    "data_to_plot = sc.Dataset()\n",
+    "N = 501\n",
+    "bins = np.linspace(2.0, 10.0, N).astype(np.float64)\n",
```
SimonHeybrock: Since yesterday, we can plot sparse data directly (binning internally), see https://scipp.readthedocs.io/en/latest/user-guide/plotting.html#Plotting-sparse-data.

dream_demo.ipynb
mads-bertelsen: The issue arises because the name can not be the same as the module name, as the cell in the notebook can not be rerun when that is the case.

dream_demo.ipynb
mads-bertelsen: It was reviewed internally without those comments, and more explanations were requested.

dream_demo.ipynb
SimonHeybrock: You will need to find a different name, or import the module under a different name (`import dream as xxx`). As soon as some `autopep8` check such as `flake8` (which we have in the `scipp` repository) is introduced, this will fail. Also, we should not teach bad style to users.

dream_demo.ipynb
mads-bertelsen: Plotting is now done using scipp histogramming and scipp plotting.

mads-bertelsen:
The dream.py module has been modified as well, the unnecessary dense Tof coords were removed. This can be done as a bug that previously required such dummy coords to be present was fixed.

dream_demo.ipynb
```diff
@@ -0,0 +1,268 @@
...
+    "print(hist_data)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "%matplotlib inline\n",
```
SimonHeybrock, line 167: This should not be needed. `scipp.plot` should setup things correctly. It works for me without, does it not for you?

mads-bertelsen: The %matplotlib inline was requested by a reviewer who couldn't run without it, I did not have issues. I suggest we leave it for wider compatibility.

SimonHeybrock: I suggest we track down the issue. How did the reviewer run it?

PR #8, opened 2019-11-13
DAS-148, porting SmoothData from Mantid to scipp

mads-bertelsen:
Resulting code from DAS-148 where the Mantid function SmoothData was ported to scipp.

SmoothData updates each point to an average of the surrounding points and itself.

The Mantid implementation uses a rolling average, where floating point errors can become problematic, especially for the errors that uses a squareroot which has to use absolute values to avoid imaginary errors.
Meanwhile the scipp version takes the mean of each point set, which is slower but both simpler and has no issues with accumulation of floating point error.

A Jupyter Notebook demo is available that shows the implementations produce equivalent results.


smoothdata.py
```diff
@@ -0,0 +1,52 @@
+import scipp as sc
+
+
+def smoothdata(variable, dim=None, NPoints=3):
```
SimonHeybrock: Consistently use snake case?

smoothdata.py
```diff
@@ -0,0 +1,52 @@
...
+            The number of points to use in the mean (odd number)
+    """
+
+    if dim is None:
+        raise ValueError("smoothdata was not given a dim to smooth.")
+
+    if NPoints < 3:
+        raise ValueError("smoothdata needs NPoints of 3 or higher.")
+
+    data_length = len(variable[dim, :].values)
```
SimonHeybrock: I am not sure what the intention here is? Are you trying to get the length in dimensions `dim`? What you are actually getting is the length of the outermost dimension (which should be what you want if the input is 1-D).

Is this what you are looking for:
```python
data_length = dict(zip(variable.dims, variable.shape))[dim]
```
This is a bit cumbersome, we should consider adding a helper to scipp that does this.

smoothdata.py
```diff
@@ -0,0 +1,52 @@
...
+
+    for index in range(hr, data_length - hr):  # Main is the bulk of the data
+        start_index = int(index - hr)
+        end_index = int(index + hr + 1)
+        out[dim, index] = sc.mean(variable[dim, start_index:end_index], dim)
+
+    for index in range(data_length - hr, data_length):  # End part
+        start_index = int(index - hr)
+        end_index = data_length  # Always use data length as endpoint
+        out[dim, index] = sc.mean(variable[dim, start_index:end_index], dim)
```
SimonHeybrock:
```python
for index in range(data_length):
    begin = max(0, index)
    end = min(data_length, index + 1)
    out[dim, index] = sc.mean(variable[dim, begin, end], dim)
```
?

SimonHeybrock (changes requested):

smoothdata.py
mads-bertelsen: I did consider using min and max, but it adds unnecessary branching. The code I ended up with should be significantly faster, but I did not benchmark.

smoothdata.py
mads-bertelsen: Agreed, will update.

smoothdata.py
SimonHeybrock: `min` and `max`of integers should be extremely fast and completely irrelevant for performance. As far as I am aware compilers do *not* generate branches from `min` and `max` (but even if they did, this is completely irrelvant on this level, have a look at what memory allocation (done by `mean` to store the output) costs compared to this).
You may even gain something because having more code is typically slower.

smoothdata.py
mads-bertelsen: I was trying to make sure it would be possible to smooth over one given dim even if several was present. Your solution certainly does what I intended. Will update.

smoothdata.py
mads-bertelsen: I timed the two solutions with 50000 points, and they are within 1% of each other as you expected. Will update to the shorter version.

mads-bertelsen:
I have pushed a new version that addresses your concerns.

SimonHeybrock (approved):

SimonHeybrock (changes requested):
You have some copies of files now.

SimonHeybrock (approved):

PR #9, opened 2019-11-20
Das 145

SimonHeybrock:


spline_background.py
```diff
@@ -0,0 +1,73 @@
...
+        variable: scipp variable
+            The variable to which the spline should be applied
+        x_values: nupmy array
+            X values for the variable data
+        dim: scipp Dim
+            The dimension along which the spline should be calculated
+    """
+
+    if dim is None:
+        raise ValueError("bspline_background: dimension must be specified.")
```
SimonHeybrock, line 24: Why default to `None` but then fail? Remove the default?

spline_background.py
```diff
@@ -0,0 +1,73 @@
...
+
+    values = variable[dim, :].values
+    errors = variable[dim, :].variances
```
SimonHeybrock: Why are you slicing here?

spline_background.py
```diff
@@ -0,0 +1,73 @@
...
+    weights = [1.0/i*i for i in errors]
```
SimonHeybrock: - Why not operate with th numpy array, avoiding the iteration and list comprehension?
- Are you sure about the square in the denominator? As the name says, `Variable.variances` are the variances, not the standard-deviations. Note that Mantid stores standard-deviations.

spline_background.py
```diff
@@ -0,0 +1,73 @@
...
+
+    # spline smoothing from
+    # https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.UnivariateSpline.html
+    s = len(values)*np.std(errors)*np.std(errors)
```
SimonHeybrock: Can you elaborate what this term is, it is not clear to me, and I cannot spot the equivalent in the Mantid algorithm?

spline_background.py
```diff
@@ -0,0 +1,73 @@
...
+    # find out the knots and splines.
+    knots, u = interpolate.splprep([x_values, values], s=s, k=5, w=weights)
+    # perform the B-spline interpolation
+    splined_variable = interpolate.splev(u, knots)
+    # splined_variable is a [2][data_length] array for X and Y values
+
+    return splined_variable
```
SimonHeybrock: Should return a `Variable` or something like that?

spline_background.py
```diff
@@ -0,0 +1,73 @@
+import numpy as np
+from scipy import interpolate
+
+
+def bspline_background(variable, x_values, dim=None):
```
SimonHeybrock: With `scipp` we can avoid the need to juggle multiple parameters. Instead of `variable` and `x_values`, accepts a `DataArray` as input, then you can get the x-values as `variable.coords[dim]`.

SimonHeybrock (commented):
@rozyczko See some comments below.

spline_background.py
rozyczko: Modified

rozyczko:
Simon's comments accepted and the code modified accordingly

spline_background.py
```diff
@@ -1,46 +1,53 @@
...
     knots, u = interpolate.splprep([x_values, values], s=s, k=5, w=weights)
     # perform the B-spline interpolation
-    splined_variable = interpolate.splev(u, knots)
-    # splined_variable is a [2][data_length] array for X and Y values
+    splined = interpolate.splev(u, knots)
+    # splined is a [2][data_length] array for X and Y values
+
+    # cast splined into DataArray type
+    output_x = sc.Variable(dims=[sc.Dim.X], values=splined[0])
+    output_y = sc.Variable(dims=[sc.Dim.X], values=splined[1])
```
SimonHeybrock: Should use input dimension instead of `Dim.X`?

spline_background.py
```diff
@@ -1,46 +1,53 @@
...
+    output_data = sc.DataArray(data=output_y, coords={sc.Dim.X: output_x})
```
SimonHeybrock: The output data has no uncertainties, is this intentional? As far as I can tell, the Mantid version does set errors: https://github.com/mantidproject/mantid/blob/13e81d9e68f084277f15dbb6bcebb8636d25f58a/Framework/CurveFitting/src/Algorithms/SplineBackground.cpp#L268-L272.

spline_background.py
rozyczko: Err.. yup. of course

spline_background.py
rozyczko: It should be included. It is now

spline_background.py
```diff
@@ -38,9 +38,9 @@ def bspline_background(variable, dim):
     # splined is a [2][data_length] array for X and Y values
 
     # cast splined into DataArray type
-    output_x = sc.Variable(dims=[sc.Dim.X], values=splined[0])
-    output_y = sc.Variable(dims=[sc.Dim.X], values=splined[1])
-    output_data = sc.DataArray(data=output_y, coords={sc.Dim.X: output_x})
+    output_x = sc.Variable(dims=[dim.X], values=splined[0])
+    output_y = sc.Variable(dims=[dim.X], values=splined[1], variances=errors)
+    output_data = sc.DataArray(data=output_y, coords={dim.X: output_x})
```
SimonHeybrock: The `.X` should also be removed, I guess?

spline_background.py
```diff
@@ -0,0 +1,86 @@
...
+    s = len(values)*np.std(errors)*np.std(errors)
+    # find out the knots and splines.
+    knots, u = interpolate.splprep([x_values, values], s=s, k=5, w=weights)
+    # perform the B-spline interpolation
+    splined = interpolate.splev(u, knots)
+    # splined is a [2][data_length] array for X and Y values
+
+    # cast splined into DataArray type
+    output_x = sc.Variable(dims=[dim.X], values=splined[0])
+    output_y = sc.Variable(dims=[dim.X], values=splined[1], variances=errors)
```
SimonHeybrock: In Mantid this output errors appear to be are based on something `gsl` does (but I did not check in detail), here you are just coping them from the input? Is gsl not modifying them?

rozyczko: splprep only returns the sum of squared residuals of the spline approximation (single number), from what I found, rather than the array of residuals.

spline_background.py
rozyczko: it should and it is now

spline_background.py
SimonHeybrock: So have you observed differences to what Mantid produces? Maybe it would be better to have an output without variances, unless you think copying the input variances is a good default?

rozyczko: gsl_multifit_wlinear() does return variances but scipy doesn't. Those of course will be different than input variances on y.
Yes, I think we should remove them altogether - they aren't correct.

SimonHeybrock: :+1:

PR #10, opened 2020-01-13
Adds ESS imaging workflow

dtasev:
This PR adds the ESS imaging workflow converted to use `scipp` for the reduction of the data. It follows the steps and uses the parameters from this [ESS reduction script](https://git.esss.dk/testbeamline/gp2/blob/master/reduce.py).

It also adds a notebook containing timings with the goal of benchmarking different steps. That notebook should not be used for reduction as it does not have all the steps performed on all the data.


scipp-imaging-2D-Bench.ipynb
```diff
@@ -0,0 +1,1585 @@
...
+    "ny_target = grouping_number\n",
+    "\n",
+    "# Number of acquired pulses per measurement\n",
+    "pulse_nr_reference = 1.0 / 770956\n",
+    "pulse_nr_sample = 1.0 / 1280381\n",
+    "pulse_nr_sample_elastic = 1.0 / 2416839\n",
+    "pulse_nr_plastic = 1.0 / 2614343\n",
+    "\n",
+    "# let's get the process started:\n",
+    "experiment_dir = \"/home/dtasev/dev/scipp_imaging/RAL_Mantid_Sep2018/\"\n",
```
OwenArnold, line 52: No hard-coded absolute paths please.

scipp-imaging-2D-Bench.ipynb
```diff
@@ -0,0 +1,1585 @@
...
+    "\n",
+    "tofs_path = os.path.join(experiment_dir, 'data_GP2', 'metadata', 'GP2_Stress_time_values.txt')\n",
+    "sample_path = os.path.join(experiment_dir, 'data_GP2', 'Stress Experiments', '2) R825')\n",
+    "sample_elastic_path = os.path.join(experiment_dir, 'data_GP2', 'Stress Experiments', '3) R825 600 Mpa')\n",
+    "ob_path = os.path.join(experiment_dir, 'data_GP2', 'Stress Experiments', '1) R825 Open Beam')\n",
+    "\n",
+    "# These use the range notation of [start, stop, step]\n",
+    "# for R825\n",
+    "frame_parameters = [(15450,22942,64),\n",
```
OwenArnold, line 61: Could you add more information here. This is more public facing that the scripts you originally worked from. What are these values, what units. Maybe could immediately be encoded as scipp variables as a self-documenting approach?

scipp-imaging-2D.ipynb
```diff
@@ -0,0 +1,8350 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import os\n",
+    "\n",
+    "# Specify where the data can be found\n",
+    "experiment_dir = \"/home/dtasev/dev/scipp_imaging/RAL_Mantid_Sep2018/\"\n",
```
OwenArnold: No absolute paths please!

scipp-imaging-2D.ipynb
```diff
@@ -0,0 +1,8350 @@
...
+    "rebin_parameters = (8500, 43000, 64)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# created by Owen Arnold (optional e-mail) on 2018-09-19\n",
```
OwenArnold: This is not correct, but you might reference the original repository and then set yourself as the author of the present script.

scipp-imaging-2D.ipynb
```diff
@@ -0,0 +1,8350 @@
...
+    "tofs_path = os.path.join(experiment_dir, 'data_GP2', 'metadata', 'GP2_BCC_time_values.txt')\n",
+    "sample_path = os.path.join(experiment_dir, 'data_GP2', 'Timeslices WFM BBC Steel')\n",
+    "ob_path = os.path.join(experiment_dir, 'data_GP2', 'Timeslices WFM Open Beam')\n",
+    "\n",
+    "# defining grouping of 2D detector pixels\n",
+    "grouping_number = 3\n",
+    "nx_target = grouping_number\n",
+    "ny_target = grouping_number\n",
+    "\n",
+    "frame_parameters = [(15167,23563,64),\n",
```
OwenArnold: Could you add more information here. This is more public facing that the scripts you originally worked from. What are these values, what units. Maybe could immediately be encoded as scipp variables as a self-documenting approach?

scipp-imaging-2D.ipynb
```diff
@@ -0,0 +1,8350 @@
...
+    "        rebinned = sc.rebin(rebinned, Dim.Tof, rebin_params)\n",
+    "\n",
+    "        frames.append(rebinned)\n",
+    "\n",
+    "    for f in frames[1:]:\n",
+    "        frames[0] += f\n",
+    "\n",
+    "    return frames[0]\n",
+    "\n",
+    "def remove_special_values(values):\n",
```
OwenArnold: Shouldn't this be `zero_special_values`?

scipp-imaging-2D.ipynb
```diff
@@ -0,0 +1,8350 @@
...
+      "GP2_007-003-WFM-open-beam_0847.tiff: Image 848, of 2001\r",
+      "GP2_007-003-WFM-open-beam_0848.tiff: Image 849, of 2001\r",
+      "GP2_007-003-WFM-open-beam_0849.tiff: Image 850, of 2001\r",
+      "GP2_007-003-WFM-open-beam_0850.tiff: Image 851, of 2001\r",
+      "GP2_007-003-WFM-open-beam_0851.tiff: Image 852, of 2001\r",
+      "GP2_007-003-WFM-open-beam_0852.tiff: Image 853, of 2001\r",
+      "GP2_007-003-WFM-open-beam_0853.tiff: Image 854, of 2001\r",
+      "GP2_007-003-WFM-open-beam_0854.tiff: Image 855, of 2001\r",
+      "GP2_007-003-WFM-open-beam_0855.tiff: Image 856, of 2001\r",
+      "GP2_007
```
OwenArnold: x_start, x_end can be moved to outer block.

OwenArnold, line 326: Question - given that the normalisation is detector by detector, is it more efficient to normalise then group or the other way around?

OwenArnold: Was it a conscious decision not to use the functions from `compat/mantid`?

OwenArnold (changes requested):

scipp-imaging-2D-Bench.ipynb
dtasev: I'm not sure how to handle this. Absolute paths aren't good, but having the relative one isn't useful either, when this notebook isn't part of the `gp2` repo. The only way it can work is to be dependent on the `gp2` repository for the `config.py` (which will be done via a

```python
from utilities import get_this_dir

```
import). Is that acceptable?

OwenArnold: I don't have to be dependent on that repository, but I think the same approach could be used. The `make_config` + products could be local to this repo.

scipp-imaging-2D.ipynb
dtasev: Good catch, `load_component_info_from_instrument_file` didn't exist (although the others did but I copied them for ease). But it is in the scipp PR now so I'll remove this block

dtasev: Good catch!

scipp-imaging-2D.ipynb
dtasev: I have renamed it as that name does make more sense.

scipp-imaging-2D.ipynb
dtasev: It is, but it changes the result values significantly from Mantid's (as division becomes the last step). 

At this point I think it's more important they stayed the same, at least until I do the value comparisons.

scipp-imaging-2D.ipynb
```diff
@@ -0,0 +1,496 @@
...
+    "        stack.append(np.flipud(img.data))\n",
+    "\n",
+    "    return np.array(stack)\n",
+    "\n",
+    "def tiffs_to_variable(tiff_dir):\n",
+    "    \"\"\"\n",
+    "    Loads all tiff images from the directory into a scipp Variable.\n",
+    "    \"\"\"\n",
+    "    stack = _load_tiffs(tiff_dir)\n",
+    "    return sc.Variable([Dim.Tof, Dim.Spectrum], \n",
```
OwenArnold, line 167: Have you tried `float32`? Maybe more relevant for the benchmark results, but selecting precision is one of the reasons for working with scipp?

scipp-imaging-2D.ipynb
```diff
@@ -0,0 +1,496 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "scrolled": true
+   },
+   "outputs": [],
+   "source": [
+    "# If you want to save the workspace uncomment this line\n",
+    "SaveNexus(ws, \"scipp_normalized_wl.nxs\")"
```
OwenArnold: `SaveNexusESS` should probably be used. Note that your `to_workspace_2d` would need to generate an instrument and attach it to output from scipp to make this fully useful.

scipp-imaging-2D.ipynb
```diff
@@ -0,0 +1,496 @@
...
+   "source": [
+    "x_coords = stitched[\"normalized_wl\"].coords[stitched[\"normalized_wl\"].dims[0]]\n",
+    "x_dim = stitched[\"normalized_wl\"].dims[0]\n",
+    "x = x_coords.values\n",
+    "\n",
+    "# Mantid expects the data in a different shape\n",
+    "# which is spectrum as outer-most dimension.\n",
+    "y = np.transpose(stitched[\"normalized_wl\"].values)\n",
+    "\n",
+    "ws = sc.compat.mantid.to_workspace_2d(x, y, None, x_dim)\n",
```
OwenArnold: Presumably the `None` here is for the `error`. Can't we propagate errors in this script? We could assign uncertainties at the point of tiff loading.

dtasev: Would they be just np.sqrt of the y values?

scipp-imaging-2D.ipynb
dtasev: I'll add the instrument as a change for https://github.com/scipp/scipp/pull/830

scipp-imaging-2D.ipynb
OwenArnold: standard deviation is sqrt counts, but variance is sq of that. So the counts again.

scipp-imaging-2D.ipynb
OwenArnold: Its unfortunate that we didn't get round to bypassing instrument 1.0 in the instrument 2.0 construction otherwise you could have created the instrument directly from your arrays.

scipp-imaging-2D.ipynb
dtasev: I have now, and I've added a new table for float32 [here](https://github.com/scipp/scipp/issues/812). It essentially doubles the speed in the slower parts, but the quick ones remain the same or very close

dtasev:
I think this is ready for another review @OwenArnold

scipp-imaging-2D.ipynb
```diff
@@ -0,0 +1,498 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# How to start\n",
+    "\n",
+    "Before starting you must:\n",
+    "- Ensure that `scipp` and `mantid` are on your `PYTHONPATH`.\n",
```
OwenArnold, line 10: These nice instructions use of config etc are not mirrored in the benchmark ipynb

dtasev: Good catch. I've updated it to point to the non-bench notebook for instructions as I don't want to duplicate those, because they will get out of date. I've also added a link to the scipp issue with previous benchmarks.

OwenArnold (approved):

PR #11, opened 2020-01-24
Completed powder_reduction.py to reproduce simple WISH workflow and r…

celinedurniak:
…enamed older version.

### DIFF
--- a/powder_reduction.py
+++ b/powder_reduction.py
@@ -6,7 +6,9 @@ from scipp import Dim
 from scipp.neutron.diffraction import load_calibration
 import smooth_data
 
-LAMBDA_RANGE = (0.7, 10.35)
+
+def get_lambda_range():
+    return 0.7, 10.35
 
 
 def powder_reduction(sample='sample.nxs', calibration=None):
@@ -43,7 +45,7 @@ def powder_reduction(sample='sample.nxs', calibration=None):
         cal = load_calibration(calibration, mantid_args=input_load_cal)
         # Merge table with detector->spectrum mapping from sample
         # (implicitly checking that detectors between sample and calibration are the same)
-        # cal = sc.merge(cal, sample.labels['detector_info'].value)
+        cal = sc.merge(cal, sample.labels['detector_info'].value)
         # Compute spectrum mask from detector mask
         mask = sc.groupby(cal['mask'], group='spectrum', combine=Dim.Spectrum).any(Dim.Detector)
 
@@ -81,7 +83,7 @@ def powder_reduction(sample='sample.nxs', calibration=None):
 
     # Rebin monitors' data
     # The value 5615 corresponds to the value in Mantid after rebinning
-    lambda_min, lambda_max = LAMBDA_RANGE
+    lambda_min, lambda_max = get_lambda_range()
     mon_rebin = sc.rebin(mon_conv,
                          Dim.Wavelength,
                          sc.Variable([Dim.Wavelength],
@@ -89,34 +91,26 @@ def powder_reduction(sample='sample.nxs', calibration=None):
                                      values=np.linspace(lambda_min, lambda_max, num=5615)))
     sample_lambda /= mon_rebin
 
-    # Delete intermediate data
     del mon_rebin, mon_conv, sample
 
-    # Convert sample data back from wavelength to tof
     sample_tof = sc.neutron.convert(sample_lambda, Dim.Wavelength, Dim.Tof)
 
-    # Delete intermediate data
     del sample_lambda
 
     # 2. Convert to d-spacing taking calibration into account
     if calibration is None:
         # No calibration data, use standard convert algorithm
-        sample_dspacing = sc.neutron.convert(data, Dim.Tof, Dim.DSpacing)
+        sample_dspacing = sc.neutron.convert(sample_tof, Dim.Tof, Dim.DSpacing)
     else:
         # Calculate dspacing from calibration file
-        # Trick to avoid error in convert_with_calibration (bug)
-        del cal['mask']
-
         sample_dspacing = sc.neutron.diffraction.convert_with_calibration(sample_tof, cal)
 
-    # Delete intermediate data
     del sample_tof
 
     # 3. Focus panels
     # Assuming sample is in Dim.DSpacing: Focus into groups
     focused = sc.groupby(sample_dspacing, group='group', combine=Dim.Group).flatten(Dim.Spectrum)
 
-    # Delete intermediate data
     del sample_dspacing
 
     # Histogram to make nice plot
@@ -126,7 +120,6 @@ def powder_reduction(sample='sample.nxs', calibration=None):
 
     focused_hist = sc.histogram(focused, dspacing_bins)
 
-    # Delete intermediate data
     del focused
 
     return focused_hist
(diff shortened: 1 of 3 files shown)


powder_reduction.py
```diff
@@ -1,37 +1,142 @@
...
     if calibration is None:
         # No calibration data, use standard convert algorithm
-        dspacing = sc.neutron.convert(data, Dim.Tof, Dim.DSpacing)
+        sample_dspacing = sc.neutron.convert(data, Dim.Tof, Dim.DSpacing)
     else:
         # Calculate dspacing from calibration file
-        dspacing = sc.neutron.diffraction.convert_with_calibration(data,
-                                                                   calibration)
+        # Trick to avoid error in convert_with_calibration (bug)
+        del cal['mask']
```
SimonHeybrock: This is fixed now so these lines can be removed.

powder_reduction.py
```diff
@@ -1,37 +1,142 @@
...
 import scipp as sc
 from scipp import Dim
+from scipp.neutron.diffraction import load_calibration
+import smooth_data
 
+LAMBDA_RANGE = (0.7, 10.35)
 
-def powder_reduction(data, calibration=None):
+
+def powder_reduction(sample='sample.nxs', calibration=None):
```
SimonHeybrock: Would this function also be valid for processing the vanadium run or are there different steps?

powder_reduction.py
```diff
@@ -1,37 +1,142 @@
+#!/usr/bin/env python
+# coding: utf-8
+import numpy as np
 import scipp as sc
 from scipp import Dim
+from scipp.neutron.diffraction import load_calibration
+import smooth_data
 
+LAMBDA_RANGE = (0.7, 10.35)
```
SimonHeybrock: Make this a function parameter (with this as the default)? Hard-coded magic values are not optimal.

powder_reduction.py
```diff
@@ -1,37 +1,142 @@
...
+
+    # Load calibration
+    if calibration is not None:
+        input_load_cal = {"InstrumentName": "WISH"}
+        cal = load_calibration(calibration, mantid_args=input_load_cal)
+        # Merge table with detector->spectrum mapping from sample
+        # (implicitly checking that detectors between sample and calibration are the same)
+        # cal = sc.merge(cal, sample.labels['detector_info'].value)
+        # Compute spectrum mask from detector mask
+        mask = sc.groupby(cal['mask'], group='spectrum', combine=Dim.Spectrum).any(Dim.Detector)
```
SimonHeybrock: I thought there is no `spectrum` column in the table. How can the `groupby` work without the `merge`?

powder_reduction.py
```diff
@@ -1,37 +1,142 @@
...
+    # 3. Focus panels
+    # Assuming sample is in Dim.DSpacing: Focus into groups
+    focused = sc.groupby(sample_dspacing, group='group', combine=Dim.Group).flatten(Dim.Spectrum)
+
+    # Delete intermediate data
+    del sample_dspacing
+
+    # Histogram to make nice plot
+    dspacing_bins = sc.Variable([Dim.DSpacing],
+                                values=np.arange(1., 10., 0.001),
```
SimonHeybrock, line 115: Why this choice of binning parameters? Is it related to the lamda range defined on top?

powder_reduction.py
```diff
@@ -1,37 +1,142 @@
...
+                         sc.Variable([Dim.Wavelength],
+                                     unit=sc.units.angstrom,
+                                     values=np.linspace(lambda_min, lambda_max, num=5615)))
+    sample_lambda /= mon_rebin
+
+    # Delete intermediate data
+    del mon_rebin, mon_conv, sample
+
+    # Convert sample data back from wavelength to tof
+    sample_tof = sc.neutron.convert(sample_lambda, Dim.Wavelength, Dim.Tof)
```
SimonHeybrock: Definitely a matter of taste, but I feel that comments like this one (and others above and below) actually decrease readability. Is the code not explicit enough?

SimonHeybrock (commented):
I think this is almost ready now (assuming that missing pieces are added as follow-ups). :+1:

Some small questions and comments below.

powder_reduction.py
celinedurniak: I just checked with the last version of scipp. It works without deleting the mask in calibration.

powder_reduction.py
celinedurniak: It is my mistake: while preparing the file, line 26 was commented out. It should not be

powder_reduction.py
celinedurniak: The choice is arbitrary. Since the boundaries of d-spacing are different for each panel of WISH, I did not want to choose niether a too small range nor a too large step size.

powder_reduction.py
celinedurniak: I removed most of the comments in the new version.

powder_reduction.py
celinedurniak: Taken into account in the new version of the script

SimonHeybrock: I did not mean a function but a function parameter to the `powder_reduction` function.

powder_reduction.py
celinedurniak: The treatment related to Vanadium could easily be added. 
For WISH, it involveds loading the corresponding files, binning and then dividing the "detectors"' dataset by the vanadium one.

powder_reduction.py
celinedurniak: OK but how are you going to proceed for the range of d-spacing, which is different for each panel?

powder_reduction.py
SimonHeybrock: My question is whether the vanadium is processed in the same way before this division, i.e., whether you can just use the existing `powder_reduction` function without changes to process the vanadium.

powder_reduction.py
SimonHeybrock: This is about the lambda range, not d-spacing? I am not sure I understand your question.

celinedurniak: I agree that your question is related to the range of wavelength. But I do not understand why you would add the range of wavelengths in the arguments of `powder reduction` and not those of d-spacing (which is group-dependent).

SimonHeybrock: Yes I would also add the latter. You could just use `scipp.Variable` to provide the group-dependent bin edges if you like. But the script under discussion does not even use group-dependent edges, so if you keep it like that you can also do without a `Variable`.

powder_reduction.py
celinedurniak: Looking at reduce.py, the correction for Vanadium looks similar to some of the steps applied to the detectors' counts. But I do not have an event file for Vanadium to run a quick test.

SimonHeybrock: Can you use the (smaller) files from the original workflow (from the Mantid systemtest)?

celinedurniak: I managed to load a nexus file for Vanadium. But it is only related to one of the panels/groups (either 5 or 6).

SimonHeybrock: I will ask for a full one. In the meantime it would be good to understand whether there are additional/different steps in the processing when handling vanadium.

celinedurniak: One step which is not included in the current workflow is the subtraction of the `empty can` data.

SimonHeybrock (approved):
Works locally for me as well. :+1:

PR #13, opened 2020-03-04
Das 182

rozyczko:
Ready for scathing review

### DIFF
--- a/absorption.py
+++ b/absorption.py
@@ -1,0 +1,141 @@
+import numpy as np
+import scipp as sc
+from scipp import Dim, Variable
+from scipp.compat.mantid import load
+import mantid.simpleapi as simpleapi
+
+def absorption_dataset(input_variable, **kwargs):
+    """
+    This method is a straightforward wrapper exposing AbsorptionCorrection through scipp
+
+    AbsorptionCorrection calculates an approximation of the attenuation due to absorption
+    and single scattering in a generic sample shape.
+
+    Requirements:
+    - The input workspace must have units of wavelength
+    - The instrument associated with the workspace must be fully defined.
+      (This being a WISH-centric implementation is done with the predefined instr file)
+    """
+
+    var_x = input_variable.coords[Dim.Tof].values
+    x_len = len(var_x)
+    var_y = [0]*x_len
+    
+    # Extract sample information
+    sample = input_variable.attrs['sample'].value
+
+    # Create workspace from passed dataset
+    ws = simpleapi.CreateWorkspace(DataX=var_x, DataY=var_y, UnitX="Wavelength")
+
+    # rebin workspace
+    #ws = simpleapi.Rebin(ws, input_variable.)
+
+    # Set sample on the workspace
+    ws.setSample(sample)
+
+    # test for shape in sample - if not there, add a default vanadium sphere
+    shape_xml = sample.getShape().getShapeXML()
+    import xml.etree.ElementTree as etree
+    root = etree.ElementTree(etree.fromstring(shape_xml)).getroot()
+    if root.text == '  ':  #'<type name="userShape">  </type>'
+
+        shape = '''<sphere id="V-sphere">
+        <centre x="0.0" y="0.0" z="0.0" />
+        <radius val="0.0025"/>
+        </sphere>'''
+
+        simpleapi.CreateSampleShape(ws, shape)
+        simpleapi.SetSampleMaterial(ws, ChemicalFormula="V")
+
+
+    # Add WISH instrument definition
+    _ = simpleapi.LoadInstrument(ws, filename="WISH_Definition.xml", RewriteSpectraMap=True)
+
+    # Run the absorption algorithm
+    ws_correction = simpleapi.AbsorptionCorrection(ws, **kwargs)
+
+    # Convert correction to Variable
+    var_correction = sc.neutron.from_mantid(ws_correction)
+
+    # The above fails with 
+    # RuntimeError: ExtractMonitors is trying to return 1 output(s) but you have provided 2 variable(s).
+    # These numbers must match.
+
+    # Apply correction
+    var_corrected = input_variable / var_correction
+
+    return var_corrected
+
+def absorption_ws(ws, **kwargs):
+    """
+    This method is a straightforward wrapper applying AbsorptionCorrection to the passed workspace
+
+    AbsorptionCorrection calculates an approximation of the attenuation due to absorption
+    and single scattering in a generic sample shape.
+
+    Requirements:
+    - The input workspace must have units of wavelength
+    - The instrument associated with the workspace must be fully defined.
+      (This being a WISH-centric implementation is done with the predefined instr file)
+    """
+
+    # Add WISH instrument definition
+    _ = simpleapi.LoadInstrument(ws, filename="WISH_Definition.xml", RewriteSpectraMap=True)
+
+    # convert to wavelength
+    ws = simpleapi.ConvertUnits(ws, Target="Wavelength")
+
+    # Run the absorption algorithm
+    ws_correction = simpleapi.AbsorptionCorrection(ws, **kwargs)
+    ws_corrected = ws / ws_correction
+
+    # Convert result to Variable
+    varOut = sc.neutron.from_mantid(ws_corrected)
+
+    return ws_corrected
+
+
+if __name__ == '__main__':
+    # simple test
+    #filename = "WISH00043525.nxs
+    #filename = "164200.nxs"
+    #filename = 'wish_test.nxs'
+    #filename = '../TrainingCourseData/PG3_4871_event.nxs'
+    filename = "164199.nxs"
+
+
+    # ------------- scipp ----------------------
+
+    # load the example file as Dataset
+    dataset = sc.compat.mantid.load(filename=filename) 
+
+      # call absorption on the scipp dataset
+    absorption_dataset1 = absorption_dataset(dataset, ScatterFrom='Sample')#, ElementSize=0.5)
+
+
+
+
+    # ------------- mantid ----------------------
+
+    # load the example file as Mantid workspace directly
+    workspace = simpleapi.Load(filename)
+    
+    # test shape
+    shape = '''<sphere id="V-sphere">
+    <centre x="0.0" y="0.0" z="0.0" />
+    <radius val="0.0025"/>
+    </sphere>'''
+
+    simpleapi.CreateSampleShape(workspace, shape)
+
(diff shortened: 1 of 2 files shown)


absorption.py
```diff
@@ -0,0 +1,88 @@
...
+    kwargs: additional arguments to be passed to Mantid's CylinderAbsorption method.
+
+    Returns
+    -------
+    Scipp dataset containing absorption correction in Wavelength units.
+
+    """
+
+    # WISH specific binning
+    lambda_binning = (0.7, 10.35, 5615)
```
SimonHeybrock: Make this a parameter to the function.

absorption.py
```diff
@@ -0,0 +1,88 @@
...
+
+    # Extract sample information
+    sample = input_variable.attrs['sample'].value
+
+    # Create empty workspace with the WISH instrument definitions
+    ws = simpleapi.LoadEmptyInstrument(Filename="WISH_Definition.xml")
+
+    # Assign the unit to X.
+    # Notice that assigning the unit to X is more complex than to Y:
+    # for Y we would have `ws.setYUnit()`
```
SimonHeybrock: Is this comment useful?

absorption.py
```diff
@@ -0,0 +1,88 @@
...
+
+    # Rebin the resulting correction based on default WISH binning
+    lambda_min, lambda_max, number_bins = lambda_binning
+    ws = simpleapi.Rebin(ws, params=[lambda_min, number_bins, lambda_max])
+
+    # Set sample on the workspace
+    ws.setSample(sample)
+
+    # Make sure a sample is defined
+    ws = ensure_sample_in_ws(ws)
```
SimonHeybrock: Is this useful? Should you rather just fail here, and let the user of `absorption_correction()` set a sample?

absorption.py
```diff
@@ -0,0 +1,88 @@
...
+
+    # Run the absorption algorithm
+    ws_correction = simpleapi.CylinderAbsorption(ws, **kwargs)
+
+    # Convert correction to DataArray
+    var_correction = sc.neutron.from_mantid(ws_correction)
+
+    # We return the dataset in wavelengths for consistency.
+    return var_correction
```
SimonHeybrock:
```suggestion
    return sc.neutron.from_mantid(ws_correction)
```

absorption.py
```diff
@@ -0,0 +1,88 @@
...
+    lambda_min, lambda_max, number_bins = lambda_binning
+    ws = simpleapi.Rebin(ws, params=[lambda_min, number_bins, lambda_max])
+
+    # Set sample on the workspace
+    ws.setSample(sample)
+
+    # Make sure a sample is defined
+    ws = ensure_sample_in_ws(ws)
+
+    # Run the absorption algorithm
```
SimonHeybrock: Probably a matter of preference, but I consider trivial comments like this and several others above as noise that hinder readability.

absorption.py
```diff
@@ -0,0 +1,88 @@
...
+    -------
+    Scipp dataset containing absorption correction in Wavelength units.
+
+    """
+
+    # WISH specific binning
+    lambda_binning = (0.7, 10.35, 5615)
+
+    # Extract sample information
+    sample = input_variable.attrs['sample'].value
```
SimonHeybrock: Having this up here and being used only further down is a bit confusing. Why not just do
```python
ws.setSample(input_variable.attrs['sample'].value)
```
below?

absorption.py
```diff
@@ -0,0 +1,88 @@
...
+        <centre-of-bottom-base x="0.0" y="0.0" z="0.0" />
+        <axis x="0.0" y="0.2" z="0.0"/>
+        <radius val="1"/>
+        <height val="10"/>
+        </cylinder>'''
+        simpleapi.CreateSampleShape(ws, shape)
+        simpleapi.SetSampleMaterial(ws, ChemicalFormula="V")
+    return ws
+
+def absorption_correction(input_variable, **kwargs):
```
SimonHeybrock: - `input_variable` -> `data` (or something like that, it is a data array, not a variable).
- For forwarding arguments to Mantid algorithms we have "standardized" on using a dict named `mantid_args`, instead of just forwarding any keyword arguments.

absorption_demo.ipynb
```diff
@@ -0,0 +1,261 @@
...
+   "outputs": [],
+   "source": [
+    "correction = absorption_correction(dataset, ScatterFrom=scatter_from)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The correction dataset uses wavelengths units, so we need to convert it back to time-of-flight, as in the original sample."
```
SimonHeybrock: Do we? Isn't the absorption applied while the data is in units of wavelength?

absorption_demo.ipynb
```diff
@@ -0,0 +1,261 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import numpy as np\n",
+    "import scipp as sc\n",
+    "from scipp import Dim\n",
```
SimonHeybrock:
```suggestion
```

absorption_demo.ipynb
```diff
@@ -0,0 +1,261 @@
...
+    "The correction dataset uses wavelengths units, so we need to convert it back to time-of-flight, as in the original sample."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "correction = sc.neutron.convert(correction, Dim.Wavelength, Dim.Tof)"
```
SimonHeybrock:
```suggestion
    "correction = sc.neutron.convert(correction, 'wavelength', 'tof')"
```

absorption_demo.ipynb
```diff
@@ -0,0 +1,261 @@
...
+    "The absorption correction has been calculated to match the input dataset over bins ranging from"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "correction.coords[Dim.Spectrum].values[0]"
```
SimonHeybrock:
```suggestion
    "correction.coords['spectrum'].values[0]"
```

absorption_demo.ipynb
```diff
@@ -0,0 +1,261 @@
...
+    "to"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "correction.coords[Dim.Spectrum].values[-1]"
```
SimonHeybrock:
```suggestion
    "correction.coords['spectrum'].values[-1]"
```

absorption_demo.ipynb
```diff
@@ -0,0 +1,261 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "correction = sc.neutron.convert(correction, Dim.Wavelength, Dim.Tof)\n",
+    "print(correction.coords[Dim.Spectrum].values[0])\n",
+    "print(correction.coords[Dim.Spectrum].values[-1])"
```
SimonHeybrock:
```suggestion
    "print(correction.coords['spectrum'].values[0])\n",
    "print(correction.coords['spectrum'].values[-1])"
```

absorption_demo.ipynb
```diff
@@ -0,0 +1,261 @@
...
+    "from scipp.plot import plot"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "plot(correction[Dim.Spectrum, :])"
```
SimonHeybrock:
```suggestion
    "plot(correction)"
```
Why slice when the full range is used?

SimonHeybrock (commented):
Looks good overall, comments are just minor improvements on conventions, readability, and recent changes to the `Dim` labels (using strings).

absorption.py
rozyczko: I left it in to remind myself about this distinction but I guess this comment is pretty redundant in general. Removed

absorption.py
rozyczko: Glad to see we agree on this. I added the comments to be overly descriptive, since this is a very short script, but of course comments should say 'why' we do something, not 'what' we do.

absorption.py
rozyczko: I added this in order for the script to run for smaller test files without sample defined.
I don't know if all data files should contain sample? If this is indeed a prerequisite then you're right - we need to throw and inform the user.

absorption.py
rozyczko: fixed

absorption.py
rozyczko: changed to `data_array`. I don't particularly like `data` as it is way too generic.

absorption.py
SimonHeybrock: Well, in practice I think almost all files contain a sample. Furthermore, the Mantid algorithm provides special parameter to define an optional different sample, if required. Overall, having a hard-coded sample does not appear useful (or will even turn out to be a reason for bugs).

If the data array has no sample, you can just skip setting it, and let Mantid handle the failure, maybe?

rozyczko: The script is now raising AttributeError on sample-less workspace.

SimonHeybrock: Yes, that is what I meant in my last sentence:
> If the data array has no sample, you can just skip setting it, and let Mantid handle the failure, maybe?

Something like
```python
if 'sample' in data.attrs:
  ws.setSample(...)
```

absorption.py
rozyczko: Fixed

absorption.py
rozyczko: Why would we want to rely on Mantid here if we can catch the failure earlier and inform the user? I would suggest
```python
if 'sample' not in data.attrs:
    raise AttributeError("Data array contains no sample.")
````

SimonHeybrock: Because, as far as I understand, if you specify certain arguments such as `CylinderSampleHeight` to the Mantid algorithm it will not require the sample.

rozyczko: I see. Fair point.

absorption_demo.ipynb
rozyczko: it is, I guess. Removed the conversion

SimonHeybrock (approved):

PR #15, opened 2020-03-05
Upgrade imaging notebooks

OwenArnold:
Objectives:

* Update scipp usage
* Introduce new plotting - PR open in scipp/core
* Benchmark for speed against existing mantid imaging - Done in benchmark notebooks
* Demonstrate to V20 instrument scientists
* Make improvements to scipp in the process

@DavidFair Edited: Updated bullets with changes


OwenArnold:
Good start. A few comments - scipp steps are loading component info information. I think this should be done up front as that's probably taking the majority of the time in some of the steps.

[25] could maybe make use of some of the zip/unzip features in `scipp.geometry` such as `position`? Then also begs the question of how up-to-date the scipp usage is.

DavidFair:
I thought I'd commit in where I'm up to (planning on going back and amending). My goal is to get it working naively then go back and tidying up / using more scipp centric constructs.

I feel like there is a couple of places the code could use scipp more optimally so if you see any feel free to point them out, such as [25] in your previous comment.

There are a couple of things outstanding:
- Does detector groupings need to be completed before converting to wavelength in Scipp? In the original it is done after the unit conversion
- Detectors are grouped together for sample / sample_elastic but not sample_plastic within the same dataset. Do we need to split into two datasets?
- Is there a mechanism to do a [weighted sum of the spectra](https://docs.mantidproject.org/nightly/algorithms/SumSpectra-v1.html) whilst converting units 
- We need some feedback since the median cannot be implemented in Scipp without finishing the variances. This was left as a TODO in the original

DavidFair:
This should be ready for review with some caveats:

I'm sure there are better ways of using SCIPP to it's full potential, but I want to take this opportunity for more experienced SCIPP devs to look at what I'm doing without changing it majorly. 

At this point I represent someone who is knowledgeable with Python, has got used to SCIPP but not aware of all the framework can offer, i.e. most users we will be onboarding in the immediate future.

If there are anti-patterns, things I'm doing fundamentally wrong or easy wins where I'm doing something in Python instead of SCIPP we should look how can we steer users to use a better alternative.

I've left the plotting as is from the original script, it could easily be improved but following the difficulty I was having with data validation I left it as-is where I could, as it's one less factor to break.

imaging/helper_funcs.py
```diff
@@ -0,0 +1,144 @@
...
+    """
+    Reads the TOF values from the CSV into a list
+    """
+    tof_values = []
+    with open(tof_file) as fh:
+        csv_reader = csv.reader(fh, delimiter='\t')
+        next(csv_reader, None)  # skip header
+        for row in csv_reader:
+            tof_values.append(float(row[1]))
+    return tof_values
```
SimonHeybrock: Can https://numpy.org/doc/stable/reference/generated/numpy.loadtxt.html do the job?

imaging/helper_funcs.py
```diff
@@ -0,0 +1,144 @@
...
+def stitch(data_array, frame_parameters, frame_shifts, rebin_parameters):
+    """
+    Stitches the 5 different frames data.
+
+    It crops out each frame, then shifts it so that all frames align,
+    and then rebins to the operations bins used for all frames.
+    """
+    frames = []
+
+    rebin_params = sc.Variable(["tof"], values=np.arange(*rebin_parameters,
```
SimonHeybrock: No unit? Usually you'd have `unit=sc.units.us` for TOF.

imaging/helper_funcs.py
```diff
@@ -0,0 +1,144 @@
...
+        bins = sc.Variable(["tof"],
+                           values=np.arange(*slice_bins, dtype=np.float64))
+        # Rebins the whole data to crop it to frame bins
+        rebinned = sc.rebin(data_array, "tof", bins)
+        # Shift the frame backwards to make all frames overlap
+        rebinned.coords["tof"] += shift_parameter
+        # Rebin to overarching coordinates so that the frame coordinates align
+        rebinned = sc.rebin(rebinned, "tof", rebin_params)
+
+        frames.append(rebinned)
```
SimonHeybrock: Can use slicing instead of rebinning for potentially much better performance?

imaging/helper_funcs.py
```diff
@@ -0,0 +1,144 @@
...
+
+        for j in range(0, ny_target):
+            y_start = j * element_width_y
+            y_end = (j + 1) * element_width_y
+
+            vals = np.full((element_width_x, element_width_y),
+                           i + j * nx_target, dtype=np.float64)
+            grid[x_start:x_end, y_start:y_end] = vals
+
+    return sc.Variable(["spectrum"], values=grid.ravel())
```
SimonHeybrock:
```python
x = sc.Variable(dims=['x'], values=np.arange(nx_original)) / element_width_x
y = sc.Variable(dims=['y'], values=np.arange(ny_original)) / element_width_y
grid = x + nx_target * y
```
or something like that?

My rule of thumb: If I have to write a look with numpy (or scipp), I am probably doing something wrong.

operations/operations.py
```diff
@@ -0,0 +1,63 @@
...
+
+def integrate(dataset: sc.Dataset):
+    integrated = sc.Dataset()
+    for dim in dataset.coords.keys():
+        integrated.coords[dim] = dataset.coords[dim]
+
+    for k in dataset.keys():
+        integrated[k] = sc.sum(dataset[k], "tof")
+
+    return integrated
```
SimonHeybrock: Why not use `sc.sum`? Why are you keeping potential coords that depending on the sum dimension, such as `tof`?

gp2-stress-experiment.ipynb
```diff
@@ -0,0 +1,933 @@
...
+    "                    (56500, 58360, 64)]\n",
+    "\n",
+    "# Used to shift the cropped frames so that their bins overlap \n",
+    "# before summing them together into a single frame\n",
+    "frame_shift_increments = [-6630, -2420, -2253, -2095, -1946, -1810]\n",
+    "frame_shift_increments = [float(i) for i in frame_shift_increments]  # Work around #1114\n",
+    "\n",
+    "# Used to rebin the summed frame in order to\n",
+    "# cut off frames that contain no data\n",
+    "rebin_parameters = (8550, 26000, (64 * 2.5))\n",
```
SimonHeybrock: Suggestion: Use a dict, because the convention (param order) in Mantid and numpy (`np.linspace`) is different.

gp2-stress-experiment.ipynb
```diff
@@ -0,0 +1,933 @@
...
+    "frame_shifts = [sum(frame_shift_increments[:i + 1]) for i in range(len(frame_shift_increments))]\n",
+    "\n",
+    "# let's get the process started:\n",
+    "tofs_path = os.path.join(data_dir, 'metadata', 'GP2_Stress_time_values.txt')\n",
+    "raw_data_dir = os.path.join(data_dir, \"Stress_Experiments\")\n",
+    "\n",
+    "ds = sc.Dataset()\n",
+    "# Load X values from the TOF file\n",
+    "ds.coords[\"tof\"] = sc.Variable([\"tof\"], unit=sc.units.us, values=imaging.read_x_values(tofs_path))\n",
+    "ds.coords[\"tof\"] *= 1e3\n",
```
SimonHeybrock: The files are in `millisecond`?

gp2-stress-experiment.ipynb
```diff
@@ -0,0 +1,933 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "scrolled": true
+   },
+   "outputs": [],
+   "source": [
+    "# Adds a coordinate for the spectra and TOF\n",
+    "ds.coords[\"spectrum\"] = sc.Variable([\"spectrum\"], values=np.arange(ds[\"sample\"].shape[1]))\n",
+    "stitched = sc.Dataset(coords={\"tof\": sc.Variable([\"tof\"], values=np.arange(*rebin_parameters, dtype=np.float64))})"
```
SimonHeybrock: No unit for TOF coord?

gp2-stress-experiment.ipynb
```diff
@@ -0,0 +1,933 @@
...
+   "source": [
+    "tof = stitched.coords['tof']\n",
+    "bin_widths = tof['tof', 1:] - tof['tof', :-1]\n",
+    "\n",
+    "integrated = sc.Dataset()\n",
+    "for dim in stitched.coords.keys():\n",
+    "    integrated.coords[dim] = stitched.coords[dim]\n",
+    "\n",
+    "for k in stitched.keys():\n",
+    "    integrated[k] = sc.sum(stitched[k], \"tof\")"
```
SimonHeybrock: `integrated = sc.sum(stitched, 'tof')`?

gp2-stress-experiment.ipynb
```diff
@@ -0,0 +1,933 @@
...
+    "            spectra_masks[i] = False\n",
+    "\n",
+    "    stitched.masks[var_name] = sc.Variable([\"spectrum\"], values=spectra_masks)\n",
+    "\n",
+    "for k in integrated.keys():\n",
+    "    if k == \"reference\":\n",
+    "        continue\n",
+    "\n",
+    "    print(f\"Masking non-sample regions in {k}\")\n",
+    "    integrated[k] = integrated[k] / integrated[\"reference\"]\n",
```
SimonHeybrock:
```suggestion
    "    integrated[k] /= integrated[\"reference\"]\n",
```
Unless you need type promotion or broadcasting, prefer in-place operation, since it is much faster.

gp2-stress-experiment.ipynb
```diff
@@ -0,0 +1,933 @@
...
+    }
+   },
+   "outputs": [],
+   "source": [
+    "def mask_non_sample_regions(integrated_dataset, var_name):\n",
+    "    integrated_spectra = integrated_dataset[var_name]\n",
+    "    spectra_masks = [False] * num_spectra\n",
+    "    for i, spec_val in enumerate(integrated_spectra.values):\n",
+    "        if spec_val > masking_threshold:\n",
+    "            spectra_masks[i] =  True\n",
```
SimonHeybrock: `spectra_mask = sc.greater(integrated_spectra, masking_threshold)`

operations/operations.py
```diff
@@ -0,0 +1,63 @@
...
+                                bank_width=bank_width,
+                                num_spectra=num_spectra),
+                        range(num_spectra))
+    assert len(results) == num_spectra
+
+    packed_pixels = {}
+    for i, adj_pixels in enumerate(results):
+        packed_pixels[i] = adj_pixels
+
+    return packed_pixels
```
SimonHeybrock: No!

```python
def shift(var, dim, forward, out_of_bounds):
  fill = var[dim, 0].copy()
  fill.values = out_of_bounds
  if forward == True:
    return sc.concatenate(fill, var[dim, :-1], dim)
  else:
    return sc.concatenate(var[dim, 1:], fill, dim)

mask = sc.reshape(spectrum_mask, dims=['y', 'x'], shape=[bank_width, bank_width])
flip = sc.Variable(dims=['neighbor', 'y', 'x'], shape=[8, bank_width, bank_width], dtype=sc.dtype.bool)
fill = True
flip['neighbor', 0] = shift(mask, x, True, fill)
flip['neighbor', 1] = shift(mask, x, False, fill)
flip['neighbor', 2] = shift(mask, y, True, fill)
flip['neighbor', 3] = shift(mask, y, False, fill)
flip['neighbor', 4:6] = shift(flip['neighbor', 0:2], y, True, fill)
flip['neighbor', 6:8] = shift(flip['neighbor', 0:2], y, False, fill)
mask = mask ^ sc.all(flip, 'neighbor')
mask = mask ^ ~sc.any(flip, 'neighbor') # if different out-of-bounds handling needed, regenerate `flip` first
# reshape mask back to 1D (or, recommended: work in XY from the start)
```
This also replaces most of the code you have in `mask_non_sample_regions`.

May need debugging, I wrote this blind.

gp2-stress-experiment.ipynb
```diff
@@ -0,0 +1,933 @@
...
+     "name": "#%%\n"
+    }
+   },
+   "outputs": [],
+   "source": [
+    "# Normalize by open beam\n",
+    "normalized = stitched.copy()\n",
+    "normalized[\"sample\"] = stitched[\"sample\"] / stitched[\"reference\"]\n",
+    "normalized[\"sample_elastic\"] = stitched[\"sample_elastic\"] / stitched[\"reference\"]\n",
+    "normalized[\"sample_plastic\"] = stitched[\"sample_plastic\"] / stitched[\"reference\"]\n",
```
SimonHeybrock: I have now seen more than one occassion indicating that having `reference` in the same dataset hurts you (requireing very verbose code). Consider making it separate, then this reduces to

```python
normalized = stitched / reference
```

gp2-stress-experiment.ipynb
```diff
@@ -0,0 +1,933 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "pycharm": {
+     "name": "#%%\n"
+    }
+   },
+   "outputs": [],
+   "source": [
+    "# Apply a median filter to the sample after converting to wavelength\n",
```
SimonHeybrock: Rewrite this without loops. Similar techniques as I used above for masking may work.

gp2-stress-experiment.ipynb
```diff
@@ -0,0 +1,933 @@
...
+   "execution_count": null,
+   "metadata": {
+    "pycharm": {
+     "name": "#%%\n"
+    }
+   },
+   "outputs": [],
+   "source": [
+    "grouped = sc.Dataset()\n",
+    "grouped[\"sample_elastic\"] = sc.groupby(rebinned_wavelength[\"sample_elastic\"], \"detector_mapping\").sum(\"spectrum\")"
```
SimonHeybrock: Without having looked further down, why use `Dataset` if there is only one entry? Use `DataArray` in that case.

SimonHeybrock (commented):
I did not look at all the details lower down yet, but here are some suggestions. Consider working with 2D images (x,y) instead of a single spectrum dimension, I think it would hlp in several places.

imaging/helper_funcs.py
DavidFair: Yeah in a surprisingly clean way, I'll keep this in mind for the future too thanks

imaging/helper_funcs.py
DavidFair: :+1: using rebin to crop instead of taking a view is really wasteful so fully agree

However, I suspect this will be a lot easier to do when we have value based slicing? Since we could say `data_array[start_tof : end_tof]` (or equivalent syntax) without having to figure out which index we need

gp2-stress-experiment.ipynb
DavidFair: This was in the original script, so I'll have to refer to @OwenArnold since he is more familiar

gp2-stress-experiment.ipynb
DavidFair: :+1: things like this are leftovers from me trying to step through the data differences we were getting

gp2-stress-experiment.ipynb
DavidFair: Shame there is no Python warning we can enable for this (since it's undetectable in the binding layer), as this is an easy mistake to make and has large performance implications

gp2-stress-experiment.ipynb
DavidFair: I'm running into a problem with this:

Trying `spectra_masks = sc.greater(integrated_spectra, sc.Variable(masking_threshold))`
Doesn't quite work since it appears that greater expects a `VariableConstView` whilst we have a `DataArrayView`, is this just a case of some missing bindings?

operations/operations.py
DavidFair: This can all be removed, it was some leftover diagnostic code trying to flush out where we were getting differences

operations/operations.py
DavidFair: I'm going to push back slightly on this, I knew it would be the controversial one haha

I agree the original approach of generating adj. spectra is not the correct way with scipp to go about this, however I wanted to unblock this review (which has been really useful)

As this will be passed back to a user I also think this code is too complex for any reasonable user; it uses shifts of the underlying data structure with bitwise mask combinations that I genuinely don't believe any scientists should be expected to support. It's just going to end up rewritten or treated as something they hope works, which is one of the grievances I get when a user passes "simple code" to a dev then gets something back they can't understand.

I'm planning on going back and adding X Y support from the start, since it will remove several places where we have to account for bank width....etc. now we have validated the data. But I'd still like to find something much more readable for users. 

Numpy has something that will do this to some extent, but I have to reach for `.values` which is a sign I'm doing something wrong:

`views = numpy.lib.stride_tricks.as_strided(var.values, shape=(3, 3), strides=bank_width)`

Which generates n number of views (that can be iterated) with the correct stride and shape of each. This is a computational pessimism as we need to iterate through views still, but CPU time is cheap, and this is going to be demo'd for buy-in that the framework make things simpler and easier to read

gp2-stress-experiment.ipynb
SimonHeybrock: Yes this is missing. Until it is available use `sc.greater(array.data(), var)`.

operations/operations.py
SimonHeybrock: > I also think this code is too complex for any reasonable user
I would argue the same about the original code. It is 3x as long and you even need to use multi-threading. Also, I gather that this is slow? How slow? Can you compare to my snippet? Users usually care surpisingly much about waiting time.

gp2-stress-experiment.ipynb
SimonHeybrock: I made this last week: https://scipp.github.io/user-guide/tips-tricks-and-anti-patterns.html.

gp2-stress-experiment.ipynb
DavidFair: Still running into issues, I've opened https://github.com/scipp/scipp/issues/1178

operations/operations.py
DavidFair: I agree completely, I'm not advocating for what we currently have it's slow and non-obvious for an end user especially the multi processing part

Using `%time` on a fresh start each time (so there is no caching of results):
Above implementation: 590ms to gen adj spectra, + ~120-150ms per data array = ~1 second

I went to benchmark the above method but it seems to segfault? I can't see anything wrong with this example (I've tweaked it to be standalone):

``` python
import scipp as sc
import numpy as np

def get_adj_pixels(mask):
    def shift(var, dim, forward, out_of_bounds):
      fill = var[dim, 0].copy()
      fill.values = out_of_bounds
      if forward == True:
        return sc.concatenate(fill, var[dim, :-1], dim)
      else:
        return sc.concatenate(var[dim, 1:], fill, dim)

    bank_width = 324

    flip = sc.Variable(dims=['neighbor', 'y', 'x'], shape=[8, bank_width, bank_width], dtype=sc.dtype.bool)
    fill = True
    flip['neighbor', 0] = shift(mask, "x", True, fill)
    flip['neighbor', 1] = shift(mask, "x", False, fill)
    flip['neighbor', 2] = shift(mask, "y", True, fill)
    flip['neighbor', 3] = shift(mask, "y", False, fill)
    flip['neighbor', 4:6] = shift(flip['neighbor', 0:2], "y", True, fill)
    flip['neighbor', 6:8] = shift(flip['neighbor', 0:2], "y", False, fill)
    mask = mask ^ sc.all(flip, 'neighbor')
    return mask ^ ~sc.any(flip, 'neighbor') # if different out-of-bounds handling needed, regenerate `flip` first

mask = sc.Variable(["y", "x"], values=np.full(shape=(324, 324), fill_value=False))
%time get_adj_pixels(mask)
```

We could document each step talking user through what each bit does, mainly how XOR is used to select which pixels to mask with some sort of ASCII diagram, but I have a deep concern about writing something that isn't user maintainable.

I'm just going to ping @OwenArnold since he was originally porting this notebook for the user, so might have a better idea on we where we should strike between simplicity and speed? If he doesn't think there will be an issue for this user then I'm happy to drop my concerns on this one and use the above.

SimonHeybrock: Ouch... we hit some bug when assigning a scalar value to `values` (I'll open an issue), do this instead:

```python
    def shift(var, dim, forward, out_of_bounds):
      fill = var[dim, 0:1].copy()
      fill.values = np.full_like(fill.values, out_of_bounds)
      # unchange below
```

```
%time 
x = get_adj_pixels(mask)
CPU times: user 4 µs, sys: 2 µs, total: 6 µs
Wall time: 13.6 µs
```

DavidFair:
I've just pushed my current local state as @OwenArnold asked to review too

@SimonHeybrock I'll put the shifts in and comment through it, as a speedup of several magnitudes this large is worth the small readability tradeoff, especially with ESS datasets. Thanks for the patience on that one

gp2-stress-experiment.ipynb
```diff
@@ -0,0 +1,949 @@
...
+    "                    # take the fitting parameters and pass them on to the forthcoming fits\n",
+    "                    params = dict(zip(sample_fit[\"parameters\"].value[\"Name\"].values, sample_fit[\"parameters\"].value[\"Value\"].values))\n",
+    "\n",
+    "                    # take the fitting parameters and pass them on to the forthcoming fits\n",
+    "                    d_sample = params[\"f1.x0\"] # See fit table definition for extract x0\n",
+    "                    fit_values = (params[\"f0.A0\"], params[\"f0.A1\"], params[\"f1.h\"], params[\"f1.a\"], params[\"f1.x0\"])\n",
+    "\n",
+    "                # Fit Bragg edge using values from fit of unstrained sample\n",
+    "                elastic_function = f\"name=LinearBackground,A0={fit_values[0]},A1={fit_values[1]};name=UserFunction,Formula=h*erf(a*(x-x0)),h={fit_values[2]},a={fit_values[3]},x0={fit_values[4]}\"\n",
+    "                fit_elastic = sc.compat.mantid.fit(elastic_ws,\n",
```
OwenArnold: Seems strange that this `fit function takes a workspace as input`. Can we create an overload that internally does the dataset->workspace conversion that you apply apove with `convert_var_to_workpace`?

gp2-stress-experiment.ipynb
```diff
@@ -0,0 +1,949 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "pycharm": {
+     "name": "#%%\n"
+    }
+   },
+   "outputs": [],
+   "source": [
+    "# Bragg edge position roughly, in Angstroem\n",
```
OwenArnold: Minor - Angstrom (probably came from original notebook)

OwenArnold (commented):
I couldn't fully run the notebooks due to the errors I reported via DM.

Aside from avoiding loops (already commented on), particularly in the applications of the spacial filter, it would be nice to re-arrange this so that the science, plotting and IO were somewhat less intertwined. I appreciate that the original scripts in mantid merged the boundaries on this and this is probably the style adopted here.

Thanks for your work so far on this. Looks promising.

operations/operations.py
DavidFair: This method wasn't working as we ended up getting a mask which looked like

```
|F|F|F|F|F
|F|T|T|T|F
|F|T|T|T|F
|F|F|F|F|F
```

When I needed basically a sliding view of each 3 by 3 for example:
```
1 | 2 | 3 | 4
5 | 6 | 7 | 8
9 | 1 | 2 | 3
```

Would return
```
1 | 2
5 | 6

1 | 2 | 3
5 | 6 | 7
```

However after running into OOM problems trying to create way to many views I stumbled on scipy Convolution2d which we can use to get the same effect

Whilst writing this I realised there is a bug, our fill value is 1. which means boundary points which get forced to `True` will work, but we need to do the convolution again for `False` with a fill value of 0. so I'll fix that later

SimonHeybrock: Fixed version in pair programming:
```python
def get_adj_pixels(mask):
    def shift(var, dim, forward, out_of_bounds):
      fill = var[dim, 0:1].copy()
      fill.values = np.full_like(fill.values, out_of_bounds)
      if forward == True:
        return sc.concatenate(fill, var[dim, :-1], dim)
      else:
        return sc.concatenate(var[dim, 1:], fill, dim)

    bank_width = 324

    def make_flip(fill):
        flip = sc.Variable(dims=['neighbor', 'y', 'x'], shape=[8, ]+mask.shape, dtype=sc.dtype.bool)
        flip['neighbor', 0] = shift(mask, "x", True, fill)
        flip['neighbor', 1] = shift(mask, "x", False, fill)
        flip['neighbor', 2] = shift(mask, "y", True, fill)
        flip['neighbor', 3] = shift(mask, "y", False, fill)
        flip['neighbor', 4:6] = shift(flip['neighbor', 0:2], "y", True, fill)
        flip['neighbor', 6:8] = shift(flip['neighbor', 0:2], "y", False, fill)
        return flip
    mask = mask | sc.all(make_flip(True), 'neighbor') # mask if all neighbors masked
    return mask & sc.any(make_flip(False), 'neighbor') # unmask if no neighbor masked
```

operations/operations.py
```diff
@@ -0,0 +1,103 @@
...
+def mask_from_adj_pixels(mask, bank_width):
+    """
+    Checks if the adjacent pixels (in 8 directions) are masked to remove
+    any noisy pixels which are erroneously masked or unmasked compared to
+    it's neighbours
+
+    If all adj. pixels are then the pixel considered is set to True
+    If no adj. pixels are then the pixel considered is set to False
+    If surrounding pixels have a mix of True/False the val is left as-is
+    This function handles border pixels as if they are "wildcard" values
```
OwenArnold: Can you clarify "wildcards"?

operations/operations.py
```diff
@@ -0,0 +1,103 @@
...
+        flip['neighbor', 6:8] = shift(flip['neighbor', 0:2], "y", False, fill)
+        return flip
+
+    # mask if all neighbors masked
+    mask = mask | sc.all(make_flip(True), 'neighbor')
+    # unmask if no neighbor masked
+    mask = mask & sc.any(make_flip(False), 'neighbor')
+
+    # Flatten using numpy, avoids #1192
+    return sc.Variable(["spectrum"], values=mask.values.ravel())
```
OwenArnold: 👍  really nice.

gp2-stress-experiment.ipynb
```diff
@@ -0,0 +1,903 @@
...
+   "execution_count": null,
+   "outputs": [],
+   "source": [
+    "def mask_non_sample_regions(integrated_dataset, var_name):\n",
+    "    integrated_spectra = integrated_dataset[var_name]\n",
+    "\n",
+    "    # This should use sc.greater once #1178 is completed\n",
+    "    spectra_masks = np.greater(integrated_spectra.values, masking_threshold)\n",
+    "    spectra_masks = sc.Variable([\"spectrum\"], values=spectra_masks)\n",
+    "    final_mask = operations.mask_from_adj_pixels(spectra_masks, bank_width)\n",
```
OwenArnold: Function above well named. Minor point: I wonder if an additional comment here explaining that this sanitises the output of the threshold operation above is necessary?

gp2-stress-experiment.ipynb
```diff
@@ -0,0 +1,903 @@
...
+    "    stitched.masks[var_name] = final_mask\n",
```
OwenArnold: There are a number of places where you access global variables like this. I don't see this as a major problem but would be nicer if this wasn't happening so much.

gp2-stress-experiment.ipynb
```diff
@@ -0,0 +1,903 @@
...
+    "indices_FCC = [(1,1,1),(2,0,0),(2,2,0),(3,1,1)]#,(2,2,2)]#,(4,0,0)]\n",
+    "indices_BCC = [(1,1,0),(2,0,0),(2,1,1),(2,2,0)]#,(3,1,0),(2,2,2)]\n",
+    "\n",
+    "def create_Braggedge_list(lattice_constant, Miller_indices):\n",
+    "    '''\n",
+    "    :param Miller_indices: like [(1,1,0),(2,0,0),...]\n",
+    "    :type Miller_indices: list of tuples\n",
+    "    '''\n",
+    "    from numpy import sqrt\n",
+    "    return [(d ,2.*lattice_constant/sqrt(d[0]**2+d[1]**2+d[2]**2)) for d in Miller_indices]\n",
```
OwenArnold: Note this is calculating the interplanar distances and then using that to calculate corresponding wavelength position Bragg condition  - nLam = 2dsin(theta). So slightly misleading to call output of comprehension `d` and should make note about return in function docs. Also why not return as a `Variable` with units Angstrom. Given Simon's refactoring of Fit, maybe that would be helpful? Or better yet, a `DataArray` with position dtype coords to store the Indices and the corresponding values stored as wavelengths?

gp2-stress-experiment.ipynb
```diff
@@ -0,0 +1,903 @@
...
+    "BCC_a = 2.865 # Angstrom, average value from COD\n",
+    "indices_FCC = [(1,1,1),(2,0,0),(2,2,0),(3,1,1)]#,(2,2,2)]#,(4,0,0)]\n",
+    "indices_BCC = [(1,1,0),(2,0,0),(2,1,1),(2,2,0)]#,(3,1,0),(2,2,2)]\n",
+    "\n",
+    "def create_Braggedge_list(lattice_constant, Miller_indices):\n",
+    "    '''\n",
+    "    :param Miller_indices: like [(1,1,0),(2,0,0),...]\n",
+    "    :type Miller_indices: list of tuples\n",
+    "    '''\n",
+    "    from numpy import sqrt\n",
```
OwenArnold: Can omit this. Already import numpy as np, just np.sqrt dirctly, may be other places that do this.

gp2-stress-experiment.ipynb
```diff
@@ -0,0 +1,903 @@
...
+   "metadata": {
+    "pycharm": {
+     "name": "#%%\n"
+    }
+   },
+   "outputs": [],
+   "source": [
+    "# Bragg edge position roughly, in Angstrom\n",
+    "FCC_a = 3.5 # Aangstrom, taken from COD entry 9008469\n",
+    "BCC_a = 2.865 # Angstrom, average value from COD\n",
```
OwenArnold: Never do anything with Ferrite BCC. Delete.

gp2-stress-experiment.ipynb
```diff
@@ -0,0 +1,903 @@
...
+    "indices_FCC = [(1,1,1),(2,0,0),(2,2,0),(3,1,1)]#,(2,2,2)]#,(4,0,0)]\n",
```
OwenArnold: I suspect that Peter has commented out the reflections that are not visible because they would fall out of the wavelength range. i.e. the bragg edges will not appear in the data. Minor point: Would be nice to verify this by calculation and made a comment to this effect. Maybe could just be clarified with Peter.

OwenArnold (commented):
Now that you've been able to significantly compress the notebook. Would you also consider putting comment/description cells in place. I think there are good examples from the docs https://scipp.github.io/tutorials/neutron-diffraction.html but obviously your focus would be on the science rather than scipp features.

gp2-stress-experiment.ipynb
DavidFair: I've added a comment to remind us to ask

gp2-stress-experiment.ipynb
DavidFair: I wrote an implementation (which I'll leave commented) however I've had to open an issue as we are currently dropping string coords, and `(1, 1, 1)` is interpreted as multiple dims rather than a single value

OwenArnold: @DavidFair I'm not following how that would be interpreted as a string. I was thinking more like ```sc.Variable(dims=['miller-indices'], values=np.array([[0,0,1], [1,1,1]]), dtype=sc.dtype.vector_3_float64) ```

gp2-stress-experiment.ipynb
OwenArnold: Let's just ask him now as he's on slack. I'll open discussion.

operations/operations.py
```diff
@@ -0,0 +1,109 @@
...
+    def shift(var, dim, forward, out_of_bounds):
+        fill = var[dim, 0:1].copy()
+        fill.values = np.full_like(fill.values, out_of_bounds)
+        if forward:
+            return sc.concatenate(fill, var[dim, :-1], dim)
+        else:
+            return sc.concatenate(var[dim, 1:], fill, dim)
+
+    mask = mask.copy()
+    mask = sc.reshape(mask, dims=["y", "x"], shape=(bank_width, bank_width))
```
SimonHeybrock: To make this function reusable (and not break as soon as you want to natively use X and Y throughout the workflow): I suggest to remove the `bank_width` parameter and deal with the reshaping externally. If you don't want it in the notebook, add a wrapping function?

operations/operations.py
```diff
@@ -0,0 +1,109 @@
...
+    # mask if all neighbors masked
+    mask = mask | sc.all(make_flip(True), 'neighbor')
+    # unmask if no neighbor masked
+    mask = mask & sc.any(make_flip(False), 'neighbor')
+
+    # Flatten using numpy, avoids #1192
+    return sc.Variable(["spectrum"], values=mask.values.ravel())
+
+
+def _calc_adj_spectra(center_spec_num, bank_width, num_spectra):
```
SimonHeybrock: Remind me why we still do this pixel-by pixel with the hack of using a thread pool?

scipp-imaging-2D.ipynb
```diff
@@ -224,218 +162,516 @@
...
    "outputs": [],
    "source": [
-    "# Load the data into the dataset\n",
-    "ds[\"sample\"] = tiffs_to_variable(sample_path)\n",
-    "ds[\"reference\"] = tiffs_to_variable(ob_path)"
+    "# Adds a coordinate for the spectra and TOF\n",
+    "ds.coords[\"spectrum\"] = sc.Variable([\"spectrum\"], values=np.arange(ds[\"sample\"].shape[1]))\n",
+    "stitched = sc.Dataset(coords={\"tof\": sc.Variable([\"tof\"], unit=sc.units.us,\n",
+    "                              values=np.arange(start=rebin_parameters[\"start\"], stop=rebin_parameters[\"stop\"],\n",
+    "                                               step=rebin_parameters[\"width\"], dtype=np.float64))})"
```
SimonHeybrock: Consider using `np.linspace`, avoiding the mess with the dtype? It lets you specify the bin count instead of the bin width, which I actually find more convenient in many cases.

scipp-imaging-2D.ipynb
```diff
@@ -224,218 +162,516 @@
...
-    "# Adds a coordinate for the spectra\n",
-    "ds.coords[Dim.Spectrum] = sc.Variable([Dim.Spectrum], values=np.arange(ds[\"sample\"].shape[1]))\n",
-    "ds"
+    "integrated = sc.Dataset()\n",
+    "for dim in stitched.coords.keys():\n",
+    "    integrated.coords[dim] = stitched.coords[dim]\n",
+    "\n",
+    "integrated = sc.sum(stitched, 'tof')\n",
+    "# Put TOF coords back for the export to TIFF stack later\n",
+    "integrated.coords[\"tof\"] = stitched.coords[\"tof\"]"
```
SimonHeybrock: A couple of things:
- First 4 lines redundant, since you just "overwrite" `integrated` with the sum.
- Last line is a bit weird to have as a coord. Have you considered other options?

scipp-imaging-2D.ipynb
```diff
@@ -224,218 +162,516 @@
...
+   "source": [
+    "# Pull out reference from out dataset to avoid checking for it in loops\n",
+    "reference = stitched[\"reference\"]\n",
+    "integrated_reference = integrated[\"reference\"]\n",
+    "\n",
+    "# Causes segmentation fault, see scipp/scipp issue: #1176\n",
+    "# when this can be added back we can remove the \"if \"reference\" checks later\n",
+    "# del stitched[\"reference\"]\n",
+    "# del integrated[\"reference\"]\n"
+   ],
```
SimonHeybrock: Why is all this needed?

scipp-imaging-2D.ipynb
```diff
@@ -224,218 +162,516 @@
...
+    "    # This should use sc.greater once #1178 is completed\n",
+    "    spectra_masks = np.greater(integrated_spectra.values, masking_threshold)\n",
+    "    spectra_masks = sc.Variable([\"spectrum\"], values=spectra_masks)\n",
+    "\n",
+    "    # Some pixels may be noisy / not strong enough signal so they erranously\n",
+    "    # become unmasked / masked from the above operation. This step checks if *all*\n",
+    "    # the surrounding pixels are True or False then sets the center pixel to\n",
+    "    # that value if they are, removing \"salt and pepper\" noise\n",
+    "    final_mask = operations.mask_from_adj_pixels(spectra_masks, bank_width)\n",
+    "    stitched_dataset.masks[var_name] = final_mask\n",
```
SimonHeybrock: This is a very Mantidy way of writing things. I suggest to `return final_mask` and change the function signature to:

```python
def mask_non_sample_regions(integrated_spectra):
```
for much more readable and understandable code on the calling site:
```python
stitched[k] = mask_non_sample_regions(integrated[k])
```

scipp-imaging-2D.ipynb
```diff
@@ -224,218 +162,516 @@
...
+    "\n",
+    "# Calculate transmission value for all\n",
+    "integrated /= integrated_reference\n",
+    "\n",
+    "for k in integrated.keys():\n",
+    "    if k == \"reference\":  # This can be removed once #1176 is resolved\n",
+    "        continue\n",
```
SimonHeybrock: Can you explain the problem? It feels like something is backwards if you are actually affected by this.

scipp-imaging-2D.ipynb
```diff
@@ -224,218 +162,516 @@
...
-   },
    "outputs": [],
    "source": [
-    "# Plot the first group\n",
-    "sc.plot.plot(sc.sum(stitched[\"normalized_grpd\"], Dim.Row)) if do_plots else None"
-   ]
+    "# Precalculate rather than doing this every loop, this will be removed in the future\n",
+    "neighbour_spectra = operations.generate_neighbouring_spectra(num_spectra=num_spectra, bank_width=bank_width)\n",
+    "\n",
+    "def median_filter(dataset, variable_key):\n",
```
SimonHeybrock: You mentioned something in the call on why you had to do this pixel-by-pixel, can you elaborate? It just looks tedious, inefficient, and most importantly error prone.

scipp-imaging-2D.ipynb
```diff
@@ -224,218 +162,516 @@
...
    "source": [
-    "# Adds the component info needed for converting units\n",
-    "sc.compat.mantid.load_component_info(stitched, instrument_file)\n",
-    "stitched"
+    "# As each detector has a different relative position to the sample we\n",
+    "# need to rebin to a common set of wavelength bins for all\n",
+    "wav_x_vals = wavelength[\"sample\"].coords[\"wavelength\"].values\n",
+    "start_wave = np.max(wav_x_vals[:, 0])\n",
+    "stop_wave = np.min(wav_x_vals[:, -1])\n",
+    "bin_width_wave = np.mean(wav_x_vals[:, 1:]-wav_x_vals[:, :-1])\n",
```
SimonHeybrock: Why not `sc.max`, `sc.min`, and `sc.mean`?

scipp-imaging-2D.ipynb
```diff
@@ -224,218 +162,516 @@
...
+    "\n",
+    "wavelength_rebin_params = sc.Variable([\"wavelength\"], values=np.arange(start_wave, stop_wave, bin_width_wave,\n",
+    "                                                           dtype=np.float64))\n",
```
SimonHeybrock: `rebin_params` is a bit misleading maybe? It is actually already the bin edges (the future coord), the params (in Mantid speak) are what you pass to `np.arange`.

scipp-imaging-2D.ipynb
```diff
@@ -224,218 +162,516 @@
...
-    "# Note: this cell is currently a workaround until \n",
-    "# https://github.com/scipp/scipp/issues/823 is done.\n",
-    "# Then it should be possible to group the positions label directly.\n",
-    "\n",
-    "# makes the position a DataConstProxy otherwise groupby won't take it\n",
-    "stitched[\"position\"] = stitched.labels[\"position\"]\n",
-    "dm1d = sc.groupby(stitched[\"position\"], \"detector_mapping\", Dim.Row)\n",
-    "position = dm1d.mean(Dim.Spectrum)\n",
+    "def weighted_sum(variable : sc.Variable, dim_to_sum):\n",
+    "    variable = variable.copy()\n",
```
SimonHeybrock: Why copy?

scipp-imaging-2D.ipynb
```diff
@@ -224,218 +162,516 @@
...
    "source": [
-    "# Note: this cell is currently a workaround until \n",
-    "# https://github.com/scipp/scipp/issues/823 is done.\n",
-    "# Then it should be possible to group the positions label directly.\n",
-    "\n",
-    "# makes the position a DataConstProxy otherwise groupby won't take it\n",
-    "stitched[\"position\"] = stitched.labels[\"position\"]\n",
-    "dm1d = sc.groupby(stitched[\"position\"], \"detector_mapping\", Dim.Row)\n",
-    "position = dm1d.mean(Dim.Spectrum)\n",
+    "def weighted_sum(variable : sc.Variable, dim_to_sum):\n",
```
SimonHeybrock: Suggest renaming `dim_to_sum` to `dim`, to match param naming in `sc.sum`.

scipp-imaging-2D.ipynb
```diff
@@ -224,218 +162,516 @@
...
    "outputs": [],
    "source": [
-    "stitched[\"normalized_wl\"] = sc.neutron.convert(stitched[\"normalized_grpd\"], Dim.Tof, Dim.Wavelength)\n",
-    "stitched"
-   ]
+    "rebinned_wavelength.coords[\"detector_mapping\"] = imaging.make_detector_groups(bank_width, bank_width,\n",
+    "                                                                              nx_target, ny_target)\n",
+    "\n",
+    "# Group detectors for the sample_elastic / sample_plastic\n",
+    "grouped = sc.groupby(rebinned_wavelength, \"detector_mapping\").sum(\"spectrum\")\n",
```
SimonHeybrock: Is this intentionally backwards from the Mantid naming? You are grouping "spectra" into "detector mappings"? I feel two other options would be a bit clearer:

- Use `'pixel'` or `'detector'` instead of `'spectrum'` everywhere. Then here: `groupby('spectrum').sum('detector')`
- Use something like `'group'` or `'spectrum_group'` here, i.e., `groupby('group').sum('spectrum')`

scipp-imaging-2D.ipynb
```diff
@@ -446,9 +682,50 @@
...
+    "    output_filenames = []\n",
+    "    for i, li in enumerate(input_list):\n",
+    "\n",
+    "\n",
+    "        bragg_edge = li[0][2].value\n",
+    "        output_filename = \"_{}_{:.3f}A_{}.txt\".format(output_filename_stem, li[0][3], bragg_edge)\n",
+    "        if not isinstance(li, list):\n",
+    "            print(\"Input is not a list.\")\n",
+    "            return\n",
+    "        whole_path = os.path.join(output_directory,output_filename)\n",
```
SimonHeybrock: https://numpy.org/doc/stable/reference/generated/numpy.savetxt.html?

operations/operations.py
DavidFair: Originally I replaced this with a median filter, however there is some nuance to the original implementation this misses which results in incorrect results.

After spending multiple days trying to nail down what is wrong with say using a median filter in SciPy or how to nicely do a 2D kernel in Numpy without using complex stride tricks I reverted back to this and left it as a TODO for another developer to come back to in a future PR as other changes weren't getting done. 

I've left this hack in with plenty of TODOs in the main notebook to replace it whilst I worked on other feedback

DavidFair: I could drop the threadpool, but I'd rather do it properly that fix a hack

scipp-imaging-2D.ipynb
DavidFair: - Good spot, from when I was originally integrating in a loop without using `sum`
- I've moved the coords copying down to the cell where we export tiffs, we TOF values which are written into a .CSV so other tools (like labview) can open the stack correctly. Copying them through seemed easiest and we don't use `integrated` after we have calculate mask positions

scipp-imaging-2D.ipynb
DavidFair: Originally it was to avoid the loops later on, but I ran into scipp/core #1176 

I've left it in so we can do things like
`integrated / integrated_reference`
`stitched / stitched_reference`

without dividing the reference by itself. Once the above issue is resolved we can uncomment it

scipp-imaging-2D.ipynb
DavidFair: Ahh yes Owen commented on the use of globals too and I went through and fixed the ones I found from the original but this slipped past :+1:

scipp-imaging-2D.ipynb
SimonHeybrock: I don't know why this is related to the issue? I can't see what is going on.

scipp-imaging-2D.ipynb
SimonHeybrock: > Copying them through seemed easiest

Coordinates in scipp are a very particular concept: They are used for ensuring correct and consistent operation. Setting random things are coords is conceptually not meaningful and will likely break some things later on. Can't you stoe it separately?

scipp-imaging-2D.ipynb
DavidFair: Once we normalize by open beam (`stitched / reference`) there is no point doing anything else with reference

However because we can't `del d["reference"]` we are forced to pay for something we don't need if we don't check for it.
If the original issue is unresolvable we could rewrite the start to keep `reference` out of the original dataset. But I'd rather keep this in as IMO it shows a user facing problem that needs to be resolved

scipp-imaging-2D.ipynb
DavidFair: I agree, see above comment ^.
I reached a point where I just burnt out looking at this dreaded function and reverted it. 

Since this produces valid data in its current state I'd rather get this PR merged then come back to it. This way the users (one of which has < 1 month left) can use this whilst we rewrite this in parallel

scipp-imaging-2D.ipynb
SimonHeybrock: > But I'd rather keep this in as IMO it shows a user facing problem that needs to be resolved

Are you referring to the bug? There is an issue for this.

The problem here is that it is very hard to understand what you are doing:

- Why do you need to add something to `d` then remove it again?
- Why do you then have to skip it?
- Why do you talk about dividing by itself above?

Isn't all that solved by just not putting it into `d`? Or if you need it there initially, just remove it, so you don't have to mess around later, just don't store a reference to it.

scipp-imaging-2D.ipynb
DavidFair: I'm trying to remove `reference` from the dataset at this point as in effect we have this happen later on if we don't

```
reference = d["reference"]
del d["reference"] # Cant do this atm
d / reference  # Divides by self so redundant 
```

scipp-imaging-2D.ipynb
DavidFair: I originally though sum worked in place so took a copy to be safe

scipp-imaging-2D.ipynb
DavidFair: > just remove it, so you don't have to mess around later, just don't store a reference to it.

But I feel like this is showing a scipp problem, I want to 

1. Do the same processing steps on all items in the dataset (including reference)
2. Take reference out of the dataset but hold onto a reference
2a) Divide all items in the dataset by reference
3. Del reference as I no longer need it from this point onwards

2a. Is the part which is "hard" since I can't do something naive like
```
ref = d["ref"]
del d["ref"]

d / ref
```

SimonHeybrock:
```python
ref = d["ref"].copy()
del d["ref"]
d / ref
```
Will eventually add `extract`, so you can do
```python
ref = d.extract('ref')
```
but that is further down the imaginary feature list.

scipp-imaging-2D.ipynb
DavidFair: No it's a product of the original Scipp notebook grouping into detector_groups using detector_mappings and the original Mantid script loading into Spectrum

I'll go through and fix up the naming

SimonHeybrock: You may need `rename_dims`:
```python
d.rename_dims({'from':'to'})
```

DavidFair: Luckily we control the loader so I renamed it to detector there, however I run into a different problem

Renaming `spectrum` to `detector` breaks the unit conversion with the following error:
`Attribute dimensions must match and not exceed dimensions of data.`

And trying to rename shows there is a dim with that name already, I've double checked and replaced all the instances of `spectrum` in this repo to no avail. 
So I've reverted it to spectrum for the time being, as I can't work on it this afternoon and Owen wants to look at it

scipp-imaging-2D.ipynb
```diff
@@ -9,433 +9,902 @@
     "Before starting you must:\n",
     "- Ensure that `scipp` and `mantid` are on your `PYTHONPATH`.\n",
     "- Generate the `config.py` file using `make_config.py`. Refer to the `README.md` or `python make_config.py --help` for information.\n",
```
OwenArnold: Is actually `scippconfig` generated by `make_config` now.

scipp-imaging-2D.ipynb
OwenArnold: Just a comment on performance - the median filter step was very slow for me.

scipp-imaging-2D.ipynb
```diff
@@ -9,433 +9,902 @@
...
-    "- Install dependency `fabio`: `conda install fabio` (used for image loading)\n",
+    "- Install dependencies : `conda install fabio tifffile` (used for image handling)\n",
     "\n",
     "For `scipp` and `mantid` follow instructions at: https://scipp.readthedocs.io/en/latest/getting-started/installation.html.\n",
     "\n",
-    "Converted to use scipp and notebook from [this file](https://git.esss.dk/testbeamline/gp2/blob/1c69213b1124982bbbe762da9c6c6457a49f2a92/reduce.py) by Dimitar Tasev on 2020-01-13"
+    "Converted to use scipp and notebook from [this repository](https://git.esss.dk/testbeamline/gp2/-/tree/stressexperiment).\n",
```
OwenArnold, line 16: Should we in addition explain what this notebook experiment is doing anyway? I think this script will certainly replace what came before so maybe increasing justification for this. Note it's finding the location of each Bragg edge as a function and using that information to measure the strain in the sample against the unstrained location. Maybe I can draft something?

scipp-imaging-2D.ipynb
```diff
@@ -446,9 +915,61 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "pycharm": {
+     "name": "#%% md\n"
+    }
+   },
+   "source": [
+    "## Generate single tile plot"
```
OwenArnold: This is running really very slowly. Been waiting for 10 minutes for it to complete. Insufficient memory maybe?

scipp-imaging-2D.ipynb
```diff
@@ -9,433 +9,902 @@
...
+    "                    spectrum_list_fitted_sample_elastic[i].append((fitted_elastic.coords[\"wavelength\"].values, fitted_elastic.values))\n",
+    "                else:\n",
+    "                    spectrum_list_fitted_sample_elastic[i].append((fitted_elastic.coords[\"wavelength\"].values,\n",
+    "                                                                                               np.zeros_like(fitted_elastic.values)))\n",
+    "                plots_counter+=1\n",
+    "\n",
+    "    return fit_list, spectrum_list_fitted_sample, spectrum_list_fitted_sample_elastic\n",
+    "\n",
+    "fit_list, spectrum_list_fitted_sample, spectrum_list_fitted_sample_elastic = \\\n",
+    "    Bragg_edge_position(xpos_Bragg_edge=Bragg_edges_FCC,\n",
```
OwenArnold, line 884: Fitting step is also quite slow. Maybe @SimonHeybrock changes will address problems here, or more to the point try again after that refactoring completed.

OwenArnold (commented):
Really nice to use except for the performance of a few steps on my laptop with just 16GB ram. Still work to do, but looking forward to running through this with user next week in current state.

operations/operations.py
```diff
@@ -0,0 +1,185 @@
...
+                            variances=has_variances)
+    container['neighbor', 0] = data
+    container['neighbor', 1] = _shift(data, "x", True, fill)
+    container['neighbor', 2] = _shift(data, "x", False, fill)
+    container['neighbor', 3] = _shift(data, "y", True, fill)
+    container['neighbor', 4] = _shift(data, "y", False, fill)
+    container['neighbor', 5:7] = _shift(container['neighbor', 1:3], "y", True,
+                                        fill)
+    container['neighbor', 7:9] = _shift(container['neighbor', 1:3], "y", False,
+                                        fill)
```
SimonHeybrock, line 88: Apart from also including the point itself, this is the same as above, and actually generic functionality. Maybe move to a common helper function, especially since writing this is slightly error prone? To overcome the slight difference: Maybe *do* include the the center point, but in the masking case slice it out in the operations, i.e., `return flip['neighbor', 1:]`?

operations/operations.py
```diff
@@ -0,0 +1,185 @@
...
+
+    edges_mask = sc.less_equal(container, sc.Variable(value=fill))
+    da = sc.DataArray(data=container, masks={'edges': edges_mask})
+    return sc.mean(da, dim='neighbor').data
+
+
+def _median(neighbors, edges_mask, dim):
+    masked_values = np.ma.array(neighbors.values,
```
SimonHeybrock, line 96: Interesting to see numpy masked arrays interacting with scipp masks.

operations/operations.py
```diff
@@ -0,0 +1,185 @@
...
+
+    def is_valid(spec_num):
+        # Do not include OOB and the centre spectra
+        return 0 <= spec_num < num_spectra and spec_num != center_spec_num
+
+    neighbour_spec = [i for i in neighbour_spec if is_valid(i)]
+    return neighbour_spec
+
+
+def generate_neighbouring_spectra(num_spectra: int, bank_width: int):
```
SimonHeybrock: This function and the one above do not seem required any more since you refactored the neighbor median code?

scipp-imaging-2D.ipynb
```diff
@@ -8,186 +8,329 @@
...
     "\n",
-    "Converted to use scipp and notebook from [this file](https://git.esss.dk/testbeamline/gp2/blob/1c69213b1124982bbbe762da9c6c6457a49f2a92/reduce.py) by Dimitar Tasev on 2020-01-13"
+    "Converted to use scipp and notebook from [this repository](https://git.esss.dk/testbeamline/gp2/-/tree/stressexperiment).\n",
+    "\n",
+    "For Table of Contents install Jupyter extensions then reload the notebook:\n",
+    "`conda install -c conda-forge jupyter_contrib_nbextensions`\n",
+    "`jupyter contrib nbextension install --user`\n",
+    "`jupyter nbextension enable toc2/main`\n",
+    "\n",
+    "# Experimental Summary\n",
```
SimonHeybrock, line 23: This section/cell appears twice?

scipp-imaging-2D.ipynb
```diff
@@ -8,186 +8,329 @@
...
    "metadata": {
-    "scrolled": true
+    "pycharm": {
+     "name": "#%%\n"
+    }
    },
    "outputs": [],
    "source": [
+    "# Lets get everything set up\n",
+    "from multiprocessing.dummy import Pool as ThreadPool\n",
```
SimonHeybrock: Not required any more?

scipp-imaging-2D.ipynb
```diff
@@ -8,186 +8,329 @@
...
-    "def tiffs_to_variable(tiff_dir):\n",
-    "    \"\"\"\n",
-    "    Loads all tiff images from the directory into a scipp Variable.\n",
-    "    \"\"\"\n",
-    "    stack = _load_tiffs(tiff_dir)\n",
-    "    data = stack.astype(np.float64).reshape(stack.shape[0], stack.shape[1]*stack.shape[2])\n",
-    "    return sc.Variable([Dim.Tof, Dim.Spectrum], \n",
-    "                       values=data, variances=data)\n",
+    "# Adds a coordinate for the spectra and TOF\n",
+    "ds.coords[\"spectrum\"] = sc.Variable([\"spectrum\"], values=np.arange(ds[\"sample\"].shape[1]))\n",
```
SimonHeybrock: Why is this coord required? It feels like it may have simple been added because "that is what Mantid does"?

scipp-imaging-2D.ipynb
```diff
@@ -196,246 +339,780 @@
...
+    "    stitched[variable_to_stitch] = stitch(ds[variable_to_stitch],\n",
+    "                                          frame_parameters=frame_parameters,\n",
+    "                                          frame_shifts=frame_shifts,\n",
+    "                                          rebin_parameters=rebin_parameters)\n",
+    "\n",
+    "# Comparing [\"sample\"][\"spectrum\", 37000] data stops matching, as Mantid\n",
+    "# and skip appear to handle to partial bin at the end differently. This results\n",
+    "# in different counts making data comparison difficult\n",
+    "for var_name in ds.keys():\n",
+    "    stitch_data(var_name)"
```
SimonHeybrock: Doesn't this work if we just call `stitch` for the whole dataset? Or is`rebin` for `Dataset` not supported yet?

scipp-imaging-2D.ipynb
```diff
@@ -196,246 +339,780 @@
...
    },
    "outputs": [],
    "source": [
-    "dm1d = sc.groupby(stitched[\"normalized\"], \"detector_mapping\", Dim.Row)\n",
-    "stitched[\"normalized_grpd\"] = dm1d.sum(Dim.Spectrum)"
+    "wav_x_vals = wavelength[\"sample\"].coords[\"wavelength\"]\n",
+    "start_wave = sc.min(wav_x_vals)\n",
+    "stop_wave = sc.max(wav_x_vals)\n",
+    "# sc.mean(wav_x_vals) throws an error, switch to using Numpy\n",
+    "bin_width = sc.Variable(unit=sc.units.angstrom, value=np.mean(wav_x_vals.values[:, 1:]-wav_x_vals.values[:, :-1]))\n",
```
SimonHeybrock: What error? Why? An issue that should be fixed? A feature? Misuse?

scipp-imaging-2D.ipynb
```diff
@@ -196,246 +339,780 @@
...
+    "    variances = sc.Variable(dims=variable.dims, values=variable.variances)\n",
+    "    d = sc.Dataset()\n",
+    "    d['data'] = variable / variances\n",
+    "    d['norm'] = sc.reciprocal(variances)\n",
+    "\n",
+    "    d.masks['zero_error'] = sc.equal(variances, 0.0 * variances.unit)\n",
+    "    \n",
+    "    d = sc.sum(d, dim)\n",
+    "    sum = d['data']/d['norm']\n",
+    "    return sum\n",
```
SimonHeybrock: Just return without defining `sum`, which is a Python keyword and may (or may not?) break things?

scipp-imaging-2D.ipynb
```diff
@@ -196,246 +339,780 @@
...
+    "pycharm": {
+     "name": "#%%\n"
+    }
    },
    "outputs": [],
    "source": [
-    "# Plot the first group\n",
-    "sc.plot.plot(sc.sum(stitched[\"normalized_grpd\"], Dim.Row)) if do_plots else None"
+    "def weighted_sum(variable : sc.Variable, dim):\n",
+    "    variances = sc.Variable(dims=variable.dims, values=variable.variances)\n",
```
SimonHeybrock: We have `sc.variances` now to make this marginally shorter.

scipp-imaging-2D.ipynb
```diff
@@ -196,246 +339,780 @@
...
-    "    return sc.Variable([Dim.Spectrum], values=grid.ravel())\n",
+    "# Load in IDF for positions of source/sample...etc\n",
+    "sc.compat.mantid.load_component_info(normalized, instrument_file)\n",
+    "wavelength = sc.neutron.convert(normalized, \"tof\", \"wavelength\")\n",
     "\n",
-    "\n",
-    "stitched.labels[\"detector_mapping\"] = make_detector_groups(324, 324, nx_target, ny_target)\n",
-    "stitched.labels[\"detector_mapping\"]"
+    "# Position data ...etc. is dropped on conversion but is later required for plotting\n",
+    "sc.compat.mantid.load_component_info(wavelength, instrument_file)"
```
SimonHeybrock: It is not dropped, it just turns into an attribute. Remove and fix below?

operations/operations.py
OwenArnold: There may be further possibilities to elide allocations, such as in-place for the `ma.median` though I don't know if that will have any large effect on performance given that we do a 9 * larger allocation anyway for the neighbours.

scipp-imaging-2D.ipynb
OwenArnold: Oh, thanks for noticing, probably messed up a merge somewhere.

OwenArnold:
Large number of changes here that should be merged. Plotting could still use some clean-up, but given the number of changes off to the side here, I'd rather that was done as a separate feature branch/PR.

PR #16, opened 2020-03-13
Das 239

rozyczko:
Simplified calculation of absorption - it's now based on loading the data file. Actual data is not read so the operation is fast.


absorption_demo.ipynb
```diff
@@ -250,24 +198,31 @@
...
-     "text": [
-      "6\n",
-      "389125\n"
-     ]
-    }
-   ],
+   "outputs": [],
    "source": [
-    "correction = sc.neutron.convert(correction, Dim.Wavelength, Dim.Tof)\n",
+    "correction = sc.neutron.convert(correction, \"Wavelength\", \"Tof\")\n",
```
SimonHeybrock: You have to use lower-case here I think, otherwise `neutron.convert` will not recognize this as known/supported dimensions.

rozyczko: Of course it should be lowercase. Fixed.

SimonHeybrock (approved):
Looks good (but you have to fix a merge conflict). 👍

PR #17, opened 2020-03-23
Added vanadium, absorption correction to workflow

celinedurniak:


powder_reduction.py
```diff
@@ -23,111 +44,459 @@ def powder_reduction(sample='sample.nxs', calibration=None, lambda_binning=(0.7,
...
+
+              number_density: number density of the sample in number of atoms
+                              per cubic Angstroms
+
+              height: height of cylindrical sample in centimeters
+
+              radius: radius of cylindrical sample in centimeters
+
+              scattering: coherent+incoherent cross section at 1.8 Angstroms
+                          in barns
```
SimonHeybrock: Why not use the sample information that comes with the file?

powder_reduction.py
```diff
@@ -23,111 +44,459 @@ def powder_reduction(sample='sample.nxs', calibration=None, lambda_binning=(0.7,
...
+    # TODO: adapt for different number of inputs
+    if bool(absorp):
+        correction = absorption_correction(sample,
+                                           lambda_binning,
+                                           ScatterFrom='Sample',
+                                           SampleNumberDensity=absorp['number_density'],
+                                           ScatteringXSection=absorp['scattering'],
+                                           AttenuationXSection=absorp['attenuation'],
+                                           CylinderSampleHeight=absorp['height'],
+                                           CylinderSampleRadius=absorp['radius'])
```
SimonHeybrock: Is this required? If the files contain a sample consider removing it to keep things simple?

powder_reduction.py
```diff
@@ -23,111 +44,459 @@ def powder_reduction(sample='sample.nxs', calibration=None, lambda_binning=(0.7,
...
+        correction = absorption_correction(sample,
+                                           lambda_binning,
+                                           ScatterFrom='Sample')
+
+    # the 5 following lines of code are to place info about source and sample
+    # position at the right place in the correction dataArray in order to
+    # proceed to the normalization
+
+    correction.attrs['source_position'] = sample_lambda.attrs['source_position'].copy()
+    correction.attrs['sample_position'] = sample_lambda.attrs['sample_position'].copy()
```
SimonHeybrock: Not needed?

powder_reduction.py
```diff
@@ -23,111 +44,459 @@ def powder_reduction(sample='sample.nxs', calibration=None, lambda_binning=(0.7,
...
+        sample_dspacing = sc.neutron.diffraction.convert_with_calibration(sample_tof, cal_sample)
+        del cal_sample
 
-    del sample_tof
-
-    # 3. Focus panels
+    # 4. Focus panels
     # Assuming sample is in d-spacing: Focus into groups
+    # deleting info in sample_dspacing as temporary fix bacause of bug in scipp
+
```
SimonHeybrock: What does this comment refer to?

powder_reduction.py
```diff
@@ -23,111 +44,459 @@ def powder_reduction(sample='sample.nxs', calibration=None, lambda_binning=(0.7,
...
+                                          posinf=replace_value,
+                                          neginf=replace_value)
+
+        variances_no_nan_inf = sc.nan_to_num(variances_to_check,
+                                             nan=replace_value,
+                                             posinf=replace_value,
+                                             neginf=replace_value)
+
+        focused_hist['group', i].values = values_no_nan_inf.values
+        focused_hist['group', i].variances = variances_no_nan_inf.values
```
SimonHeybrock: Why are you doing this group by group?

Furthermore, I think setting values and variances to 0 should lead to bad results. Shouldn't bad values be masked instead?

powder_reduction.py
```diff
@@ -23,111 +44,459 @@ def powder_reduction(sample='sample.nxs', calibration=None, lambda_binning=(0.7,
...
+
+    # Sum final result for plotting
+    sample_sum = sc.sum(focused_hist, 'group')
+
+    return sample_sum
```
SimonHeybrock: Return event data, so the user of the function can decide whether they want to histogram or not?

powder_reduction.py
```diff
@@ -23,111 +44,459 @@ def powder_reduction(sample='sample.nxs', calibration=None, lambda_binning=(0.7,
...
+
+    # ################################
+    # Monitor correction
+    # extract monitor
+    mon4_selected = event_data.attrs['monitor4'].values
+
+    # convert to wavelength
+    mon4_lambda = sc.neutron.convert(mon4_selected, 'tof', 'wavelength')
+
+    del mon4_selected
```
SimonHeybrock: Shorter?
```suggestion
    mon4_lambda = sc.neutron.convert(event_data.attrs['monitor4'].value, 'tof', 'wavelength')
```

powder_reduction.py
```diff
@@ -23,111 +44,459 @@ def powder_reduction(sample='sample.nxs', calibration=None, lambda_binning=(0.7,
...
+
+    # spline_background
+    # vana_mon4_spline = bspline_background(vana_mon4_lambda,
+    #                                       sc.Dim('wavelength'),
+    #                                       smoothing_factor=70)
+    # del vana_mon4_lambda
+    # smooth data
+    # vana_mon4_smooth = smooth_data(vana_mon4_spline, dim='wavelength', NPoints=40)
+    # del vana_mon4_spline
```
SimonHeybrock: Commented code?

powder_reduction.py
```diff
@@ -23,111 +44,459 @@ def powder_reduction(sample='sample.nxs', calibration=None, lambda_binning=(0.7,
...
+    if bool(absorp):
+        correction = absorption_correction(vanadium,
+                                           lambda_binning,
+                                           ScatterFrom='Sample',
+                                           SampleNumberDensity=absorp['number_density'],
+                                           ScatteringXSection=absorp['scattering'],
+                                           AttenuationXSection=absorp['attenuation'],
+                                           CylinderSampleHeight=absorp['height'],
+                                           CylinderSampleRadius=absorp['radius'])
+    else:
```
SimonHeybrock: See above.

powder_reduction.py
```diff
@@ -23,111 +44,459 @@ def powder_reduction(sample='sample.nxs', calibration=None, lambda_binning=(0.7,
...
+        correction = absorption_correction(vanadium,
+                                           lambda_binning,
+                                           ScatterFrom='Sample')
+
+    # the 5 following lines of code are to place info about source and sample
+    # position at the right place in the correction dataArray in order to
+    # proceed to the normalization
+
+    correction.attrs['source_position'] = vana_red.attrs['source_position'].copy()
+    correction.attrs['sample_position'] = vana_red.attrs['sample_position'].copy()
```
SimonHeybrock: Not necessary?

powder_reduction.py
```diff
@@ -23,111 +44,459 @@ def powder_reduction(sample='sample.nxs', calibration=None, lambda_binning=(0.7,
...
+    vana_rebin = sc.rebin(vana_dspacing, 'd-spacing', dspacing_bins)
 
-    focused_hist = sc.histogram(focused, dspacing_bins)
+    del vana_dspacing
 
-    del focused
+    # Calibration
+    # Load
+    input_load_cal = {'InstrumentFilename': 'WISH_Definition.xml'}
+    calvana = load_calibration(calibration, mantid_args=input_load_cal)
```
SimonHeybrock, line 375: Does this work properly now? When I ran this last week Mantid was really slow and had apparently trouble with tons of bad detector IDs.

powder_reduction.py
```diff
@@ -23,111 +44,459 @@ def powder_reduction(sample='sample.nxs', calibration=None, lambda_binning=(0.7,
...
-                                    lambda_binning=(0.7, 10.35, 5615))
+    vana_rebin.coords['group'] = groupvana.data
+    vana_rebin.masks['mask'] = maskvana.data
+
+    # Mask negative values in measured values
+    for i in range(vana_rebin.coords['spectrum'].shape[0]):
+        single_spectrum = vana_rebin['spectrum', i].values
+        detect_if_negative_values = any(item < 0 for item in single_spectrum)
+        if detect_if_negative_values:
+            vana_rebin.masks['mask'].values[i] = True
```
SimonHeybrock: So you are masking the complete spectrum even if there is just a single negative bin? Why not masks bin-by-bin?

```python
vana_rebin.masks['negative_bins'] = sc.Variable(dims=..., values=np.less(vana_rebin.values, 0.0))
```
```

File: powder_reduction.py [hunk 13]
```diff
@@ -23,111 +44,459 @@ def powder_reduction(sample='sample.nxs', calibration=None, lambda_binning=(0.7,
...
+
+    # focus (deleting info in vana_dspacing because of bug in scipp)
```
SimonHeybrock: What is this referring to?

powder_reduction.py
```diff
@@ -23,111 +44,459 @@ def powder_reduction(sample='sample.nxs', calibration=None, lambda_binning=(0.7,
...
 
-    plot(focused_hist)
+    # #spline_background
+    # vana_focus_spline = bspline_background(vana_dspacing,
+    #                                        sc.Dim('d-spacing'),
+    #                                        smoothing_factor=120)
+    # del vana_dspacing
+
+    # #smooth_data
+    # smooth_data(vana_focus_spline, dim='d-spacing', NPoints=40)
```
SimonHeybrock: Commented code?

powder_reduction.py
```diff
@@ -23,111 +44,459 @@ def powder_reduction(sample='sample.nxs', calibration=None, lambda_binning=(0.7,
...
+
+if __name__ == "__main__":
+
+    # The value 5615 for the number of bins corresponds to the value in Mantid after rebinning
+    # No vanadium
+    # focused_hist = powder_reduction(sample='WISH00043525.nxs',
+    #                                 calibration="WISH_cycle_15_4_noends_10to10_dodgytube_removed_feb2016.cal",
+    #                                 vanadium=None,
+    #                                 empty_instr=None,
+    #                                 lambda_binning=(0.7, 10.35, 5615))
```
SimonHeybrock: Commented code?

spline_background.py
```diff
@@ -32,57 +35,45 @@ def bspline_background(variable, dim, smoothing_factor=None):
...
+    bin_edge_input = False
+
+    # If input x values are bin-edges, use bin centres to spline
+
+    if len(x_values) == len(values)+1:
+        bin_edge_input = True
+
+        new_x_values = [0.5*(x_values[i] + x_values[i+1])
+                        for i in range(len(x_values)-1)]
+        x_values = new_x_values
```
SimonHeybrock:
```suggestion
        x_values = 0.5*(x_values[:-1] + x_values[1:])
```

spline_background.py
```diff
@@ -32,57 +35,45 @@ def bspline_background(variable, dim, smoothing_factor=None):
...
+                                         3*splined[0][len_splined-1] -
+                                         splined[0][len_splined-2])
+
+        for i in range(len_splined-1):
+            bin_edge_output_x[i+1] = 0.5*(splined[0][i+1] - splined[0][i])
+
+        output_x = sc.Variable(dims=[dim], values=bin_edge_output_x)
+    else:
+        output_x = sc.Variable(dims=[dim], values=splined[0])
+
```
SimonHeybrock:
```suggestion
    output_x = variable.coords[dim]
```

spline_background.py
```diff
@@ -32,57 +35,45 @@ def bspline_background(variable, dim, smoothing_factor=None):
     errors = variable.variances
     weights = 1.0/errors
 
-    # https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.UnivariateSpline.html
+    # flag marking 'bin-edge' status of input x values
+    bin_edge_input = False
```
SimonHeybrock, line 39: Not needed anymore (see common on using input X as output X).

powder_reduction.py
celinedurniak: I removed the line of code but forgot to remove the comment above it.

powder_reduction.py
celinedurniak: These steps are in the initial WISH workflow. But since I had issues with applying spline_background, I commented these 2 lines out.

powder_reduction.py
celinedurniak: It was the old implementation without absorption. I'll remove this part.

powder_reduction.py
SimonHeybrock: This is not clear to the reader. Commented code should be avoided since it makes the purpose unclear:
- Is there a bug?
- Has this been accidentially disabled?
- Is it broken?
- Is it still needed?
- Harder to read?
- `git` remembers code, why not remove it.

celinedurniak: OK I'll remove these lines here and also from line 283.

powder_reduction.py
celinedurniak: I cropped the details about **absorp to

**absorp: dictionary containing information to correct absorption  
see help of Mantid's algorithm CylinderAbsorption for details   https://docs.mantidproject.org/nightly/algorithms/CylinderAbsorption-v1.html

celinedurniak: Which file are you talking about?

SimonHeybrock: The NeXus files.

powder_reduction.py
celinedurniak: I just checked. These 2 lines are not needed anymore. They'll be removed.

powder_reduction.py
celinedurniak: The minimum number of arguments is sample, lambda_binning, CylinderSampleHeight and CylinderSampleRadius.

SimonHeybrock: Shouldn't the file contain the sample information such as height and radius?

celinedurniak: I just loaded the Nexus file for the sample, checked the sample logs and sample material. There is no info about the height, radius of the sample shape.

Screen Shot 2020-03-23 at 14 15 42

SimonHeybrock: Does the WISH workflow run absorption correction for the sample? Maybe just for vanadium?

celinedurniak: I just checked reduce.py. I think the absorption is also applied to the sample data.
I also loaded the Vanadium nexus file into Mantid. Same thing: no info about the sample in Sample material.

powder_reduction.py
celinedurniak: OK I'll merge the 2 lines here and for the Sample data as well

powder_reduction.py
SimonHeybrock: Surprising (how do the users know the sizes for any given sample?), but I suppose we have to leave it in then. 👍

powder_reduction.py
celinedurniak: I'll remove these lines

powder_reduction.py
celinedurniak: I cropped the input arguments to the minimum required

SimonHeybrock: Have you considered using a `dict`, as scipp does with the `mantid_args` for `LoadEventNexus`?

powder_reduction.py
celinedurniak: Will be removed.

powder_reduction.py
celinedurniak: I had the same issues: slow and message about bad detectors. But it is the only way I found to get the groups for the correct amount of spectrum. 
This issue is due to the fact that Sample and Vanadium data were not recorded using the same layout of detectors. In normal cases, it should not be the case.

powder_reduction.py
celinedurniak: So you would remove from 6.). Am I correct?

SimonHeybrock: I think so. I suppose you can move the code to an independent helper function for convenience, but not call it here.

celinedurniak: OK. So one of the points (replacement of Nan, inf, -inf) is not relevant anymore.

powder_reduction.py
celinedurniak: This part of the script has been removed in order to get events as output.

powder_reduction.py
celinedurniak: You mean line 463 only?

SimonHeybrock: Yes.

spline_background.py
celinedurniak: I modify the script to this new version

spline_background.py
celinedurniak: The input x-values cannot be simply copied to the output because the scipy algorithms modify it.

powder_reduction.py
celinedurniak: I modified the script accordingly.

spline_background.py
SimonHeybrock: Ah, ok. You can still shorten by doing parts of the math with numpy arrays though.

powder_reduction.py
```diff
@@ -23,111 +44,346 @@ def powder_reduction(sample='sample.nxs', calibration=None, lambda_binning=(0.7,
...
+                                     values=np.linspace(lambda_min,
+                                                        lambda_max,
+                                                        num=number_bins)))
     sample_lambda /= mon_rebin
 
-    del mon_rebin, mon_conv, sample
+    del mon_rebin, mon4_smooth
+
+    # 2. absorption correction
+    # TODO: adapt for different number of inputs
```
SimonHeybrock: Why not just unpack the dict (dict keys have to match Mantid argument names) below:
```
bsorption_correction(sample, lambda_binning, **absorp)
```
Then there is no "TODO" and it should be flexible and keep working irrespective of future changes?

SimonHeybrock: Or maybe even better: Just pass the dict to `absorption_correction` and let that do the unpacking?

powder_reduction.py
```diff
@@ -3,14 +3,35 @@
...
 
 
-def powder_reduction(sample='sample.nxs', calibration=None, lambda_binning=(0.7, 10.35, 5615)):
+def powder_reduction(sample='sample.nxs',
+                     calibration=None,
+                     vanadium=None,
+                     empty_instr=None,
+                     lambda_binning=(0.7, 10.35, 5615),
+                     dspacing_binning=(1., 10., 9000),
+                     **absorp):
```
SimonHeybrock, line 17: This implies that the same sample size is used for sample and vanadium runs. Is guess this is not true in practice, i.e., you should probably have independent absorption parameters for the two?

powder_reduction.py
celinedurniak: I just submitted a new version of the file to deal with different values for the absorption correction related to Sample and Vanadium.

PR #18, opened 2020-05-04
Updated powder_reduction (realigned)

celinedurniak:


powder_reduction.py
```diff
@@ -163,9 +168,11 @@ def powder_reduction(sample='sample.nxs',
         del correction.coords['sample_position']
         del correction.coords['position']
 
-        sample_lambda /= correction
+        correction_rebin = sc.rebin(correction, 'wavelength', edges_lambda)
 
-        del correction
+        sample_lambda /= correction_rebin
+
+        del correction, correction_rebin
```
SimonHeybrock: Why remove these explicitly? Don't they go out of scope at the end of the `if`?

powder_reduction.py
```diff
@@ -186,7 +193,7 @@ def powder_reduction(sample='sample.nxs',
     # 4. Focus panels
     # Assuming sample is in d-spacing: Focus into groups
 
-    focused = sc.groupby(sample_dspacing, group='group').flatten('spectrum')
+    focused = sc.groupby(sample_dspacing.unaligned, group='group').flatten('spectrum')
```
SimonHeybrock: Does
```python
focused = sc.groupby(sample_dspacing, group='group').sum('spectrum')
```
work here? It should in principle, but I have not tested this in all cases yet, so I cannot exclude that it does not.

powder_reduction.py
```diff
@@ -208,7 +215,18 @@ def powder_reduction(sample='sample.nxs',
 
         vana_red_focused.coords['detector_info'] = focused.coords['detector_info'].copy()
 
-        focused /= vana_red_focused
+        # realign vanadium and sample data in d-space
```
SimonHeybrock: If we use `groupby` on the realigned data (see previous comment), is this step necessary? The realignment should in principle carry through all operations (if it does not this is something we need to fix), it should only have to be done once.

powder_reduction.py
```diff
@@ -360,22 +377,22 @@ def process_vanadium_data(vanadium, empty_instr,
...
-    vana_rebin = sc.rebin(vana_dspacing, 'd-spacing', dspacing_bins)
+    # # rebin vanadium
+    # dmin, dmax, number_bins_d = dspacing_binning
+    # dspacing_bins = sc.Variable(['d-spacing'],
+    #                            values=np.linspace(dmin, dmax, number_bins_d),
+    #                            unit=sc.units.angstrom)
+    # vana_rebin = sc.rebin(vana_dspacing, 'd-spacing', dspacing_bins)
 
-    del vana_dspacing
+    # del vana_dspacing
```
SimonHeybrock: Leftover commented code here and above.

powder_reduction.py
```diff
@@ -386,17 +403,17 @@ def process_vanadium_data(vanadium, empty_instr,
...
-    vana_rebin.masks['mask'] = maskvana.data
+    vana_dspacing.coords['group'] = groupvana.data
+    vana_dspacing.masks['mask'] = maskvana.data
 
-    # Mask negative measured values
-    vana_rebin.masks['negative_bins'] = sc.Variable(dims=['spectrum', 'd-spacing'],
-                                                    values=np.less(vana_rebin.values, 0.0))
+    # # Mask negative measured values
+    # vana_rebin.masks['negative_bins'] = sc.Variable(dims=['spectrum', 'd-spacing'],
+    #                                                 values=np.less(vana_rebin.values, 0.0))
```
SimonHeybrock: More leftover commented code.

powder_reduction.py
```diff
@@ -386,17 +403,17 @@ def process_vanadium_data(vanadium, empty_instr,
...
 
     # Focus
-    focused_vana = sc.groupby(vana_rebin, group='group').flatten('spectrum')
+    focused_vana = sc.groupby(vana_dspacing.unaligned, group='group').flatten('spectrum')
```
SimonHeybrock: See previous comment, should be able to use `groupby(..).sum(..)` instead of going back to unaligned data.

powder_reduction.py
celinedurniak: I removed the commented code.

powder_reduction.py
celinedurniak: I removed the commented code.

powder_reduction.py
celinedurniak: It will be removed

powder_reduction.py
celinedurniak: It works without "unaligned". But there is no data.

powder_reduction.py
celinedurniak: See above comment

powder_reduction.py
SimonHeybrock: Correct, `focused` would also be realigned data in this case, but that is what you need, I believe?

celinedurniak: You are correct. At this final stage, it is what I need.

celinedurniak:
My last commit also modified spline_background.py since there was an error with the new version of scipp because of mismatching units between input and output.

SimonHeybrock (approved):

PR #19, opened 2020-05-12
First draft of reduction script

SimonHeybrock:
See comments and TODOs in script.


sans/reduction.ipynb
```diff
@@ -0,0 +1,400 @@
...
+    "def extract_monitor_background(data, begin, end):\n",
+    "    background = background_mean(data, 'tof', begin, end)\n",
+    "    del background.coords['sample_position'] # ensure unit conversion treats this a monitor\n",
+    "    background = sc.neutron.convert(background, 'tof', 'wavelength')['spectrum', 0]\n",
+    "    background = sc.rebin(background, 'wavelength', wavelength_bins)\n",
+    "    return background\n",
+    "\n",
+    "def setup_transmission(data):\n",
+    "    incident_beam = extract_monitor_background(data['spectrum', 0:1], 40000.0*sc.units.us, 99000.0*sc.units.us)\n",
+    "    transmission = extract_monitor_background(data['spectrum', 3:4], 88000.0*sc.units.us, 98000.0*sc.units.us)\n",
```
SimonHeybrock, line 198: @OwenArnold @nvaytet Note the hack here to use range-based slice `['spectrum', 0:1]`. What we actually want is `['spectrum', 0]`, but this drops the position coord and unit conversion will not work any more. What is the best way to address this, here and in general? Should we store a workspace containing only monitors as a dataset instead, so this changes from:
```python
data['spectrum', 0:1]
```
to
```python
data['monitor17']
```
?

SimonHeybrock: Actually that does not work, because the positions coords of the monitors conflict when added to a `Dataset`.

SimonHeybrock: Maybe it should simply be a dict of data arrays?

SimonHeybrock: @OwenArnold @nvaytet The issue with the requirement to use range-slicing here is interesting, because it highlights a difference between xarray and scipp, and may point at a potential solution:

**xarray**
- Slicing without range in xarray *preserves* all coordinates (even those of dims that are sliced, they become scalar).
- Modifications via non-range slices works because xarray *compares only dimension-coords* in operations.

**scipp**
- Slicing without range *drops* coords along dims that are sliced.
- Modification via non-range slices works because (despite scipp comparing *all coords, including non-dimension-coords) there are no conflicting coords.

**Question**

What is the right approach? Obviously both work. The question is: Do we **need** comparison of non-dimension coords, i.e., is this a valuable safety net, or more of a hindrance (causing problems such as dropped position coord when extracting a single spectrum via slicing)?

OwenArnold: From the sounds of it, the `scipp` approach seems like the correct one to me. What we would be doing otherwise is still having coords with some special case loss of integrity rules for non-dimension-coords?

I can't see the advantage of preserving the coord as a scalar when non-range slicing? Maybe this is useful for diagnostics down the line?

SimonHeybrock: > I can't see the advantage of preserving the coord as a scalar when non-range slicing? Maybe this is useful for diagnostics down the line?

By far more than diagnostics. For example, to extract a single spectrum. Right now you lose the position information and will not be able to run unit conversion any more.

SimonHeybrock: Have a look at http://xarray.pydata.org/en/stable/computation.html#coordinates. They have three examples there:

```python
# slicing x
arr[1] - arr[0] # no x coord
arr[0] + 1 # has x coord
arr[0] - arr[0] # has x coord,
```

In scipp, none of these have an x coord. The maybe more relevant case is the reverse though: Consider a 1D data array that represents a TOF monitor. It should have a `position` coord, such that unit conversion works. This is not a dimension coord, because `position` is not a dimension of the monitor. Scipp will however prevent binary operations in this case --- users are forced to change the coord to an attribute.

Now, I am wondering if non-dimension coords in xarray behave almost like attributes in scipp?

Writing this, I am wondering if this could solve some of our issues with unit conversions: The problem was that we have to switch, e.g., positions from coord to attributes when doing a conversion to wavelength, such that we can then normalize to a monitor (which would have a conflicting position):
```python
# data has dims=('spectrum', 'tof'), `position` is a non-dimension coord
data = sc.neutron.convert(data, 'tof', 'wavelength') # weird coord->attr juggling, partially broken
```

 Now, with xarrays behavior for non-dimension coords, our conversion could be:
```python
# data has dims=('position', 'tof'), `position` is a dimension coord
data = sc.neutron.convert(data, {'source-pos', 'sample-pos', 'position', 'tof'}, {'theta', 'wavelength'})
# data has dims=('theta', 'wavelength')
# new coord `theta` is dimension coord
# `position` is a non-dimension coord
```
Ignoring things like `source-position` and `sample-position` coords, this appears to solve some (not all) of the issues we had:
- No weird/surprising changes between coords and attributes.
- Normalizing to monitor works, since position is not a dimension coord.
- Theta (needed for `Q` later) is naturally there, no need to compute it based on position *attribute*.

Downsides/unclear:
- Adding, e.g., data and monitor to the same dataset will drop scalar position coord of monitor (if data is added first), `dataset['monitor']` will not show position.

PR #20, opened 2020-05-13
Make easier to work locally

OwenArnold:
Prevent conflicts every time we edit to work locally


sans/reduction.ipynb
```diff
@@ -45,7 +56,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "path = '/home/simon/code/loki_tube_scripts/test/test_data'\n",
+    "path = os.path.join(dataconfig.data_root,'test_data')\n",
```
SimonHeybrock:
```suggestion
    "path = dataconfig.data_root\n",
```
Will want to run this also with files from elsewhere.

PR #21, opened 2020-05-18
Playing with potential unit conversion API

SimonHeybrock:
A first attempt (not for merging).

Problems:
- Data that is already in wavelength ends up with "non-compliant" position coord (which should be an attribute), since we never call `TofBeamline.convert` on it (see TODO in code). Ideas on how to solve this nicely?


sans/reduction.ipynb
```diff
@@ -236,12 +247,11 @@
     "    # Setup direct beam and normalise to monitor. I.e. adjust for efficiency of detector across the wavelengths.\n",
     "    direct_beam = load_rkh(filename=f'{path}/{direct_beam_file}')\n",
     "    del direct_beam.coords['detector_info']\n",
-    "    del direct_beam.coords['position']\n",
     "    to_bin_edges(direct_beam, 'wavelength')\n",
     "    direct_beam = rebin_distribution(direct_beam, 'wavelength', monitor.coords['wavelength'])\n",
     "    direct_beam.unit = sc.units.one # TODO conflicts with unit conv to Q, is Q1D2 ignoring this?\n",
     "    direct_beam = monitor * transmission * direct_beam\n",
-    "    del direct_beam.coords['position']\n",
+    "    del direct_beam.coords['position'] # TODO already in wavelength -> never using TofBeamline -> non-standard coords\n",
```
SimonHeybrock, line 254: How can we avoid this elegantly? Is it something to be fixed in the conversion from Mantid?

sans/reduction.ipynb
```diff
@@ -177,6 +177,18 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "beamline = sc.neutron.TofBeamline(\n",
+    "    path=['source_position', 'sample_position', 'position'],\n",
+    "    scatter='sample_position')\n",
+    "elastic = sc.neutron.ElasticScattering(theta='theta');"
```
SimonHeybrock, line 189: Here we could add extra options (or use different converters) to support things such as gravity.

sans/reduction.ipynb
SimonHeybrock: To clarify: I am thinking of something like making positions coords only if the workspace contains time-of-flight. If it is another unit, positions would be attributes.

PR #23, opened 2020-05-22
loki q resolution

OwenArnold:
Based on https://docs.mantidproject.org/nightly/algorithms/TOFSANSResolutionByPixel-v1.html see also https://github.com/ess-dmsc-dram/loki_tube_scripts/blob/master/data_reduction_scripts/FullDataReduction.py#L201-L208

There is a corresponding branch [loki_q_resolution](https://github.com/scipp/scipp/pull/1163) for the scipp repository 

----------
Notes of Combining q1d and q_resolution steps based on Mantid Q1D

* qres and input data workspace have guaranteed same wavelength bins in mantid, `Qhelper` checks this. 
* Qout is binning input to algorithm (already handled in scipp)
* Qout [bin] is lookup from Qin. Qin is calculated from input wavelengths via simple transform per bin.
* QBin is bin width in output Q
* QResIn is q-value at corresponding wavelength.

[here](https://github.com/mantidproject/mantid/blob/bb3320265bfd15f6a9820816e12e5548d1127930/Framework/Algorithms/src/Q1D2.cpp#L257)

```
qResolutionOut[bin] +=
                (*YIn) * std::sqrt((*QResIn) * (*QResIn) + QBin * QBin / 12.0);
```
I think this could be done via `histogram`. With input being ```d * sqrt(dq * dq)``` followed by addition of ```1/12* Qbin_width ** 2```

[here](https://github.com/mantidproject/mantid/blob/bb3320265bfd15f6a9820816e12e5548d1127930/Framework/Algorithms/src/Q1D2.cpp#L326)

```
*qResolutionIterator = (*qResolutionIterator) / (*countsIterator);
```
[here](https://github.com/mantidproject/mantid/blob/bb3320265bfd15f6a9820816e12e5548d1127930/Framework/Algorithms/src/Q1D2.cpp#L329)
```
outputWS->setPointStandardDeviations(0, std::move(qResolutionOut));
```


sans/reduction.ipynb
```diff
@@ -210,6 +210,55 @@
...
+    "    moderator.data.unit = sc.units.s # fixup moderator units to microseconds. Previously dimensionless\n",
+    "    \n",
+    "    # Calcualte moderator spread\n",
+    "    s = moderator.data * sc.reciprocal(l1 + l2) # times to inv velocity\n",
+    "    s *= k # inv velocity to wavelength\n",
+    "    s.unit = sc.units.angstrom\n",
+    "    s *= 1e10\n",
+    "    \n",
+    "    std_dev_lam = (d_lam)/12 + s\n",
+    "    std_dev_lam *= sc.reciprocal(lam)\n",
```
SimonHeybrock: Missing squares on `d_lam`, `s`, and `lam` (or `sqrt` on 12)?

sans/reduction.ipynb
```diff
@@ -210,6 +210,55 @@
...
+    "    f = (4 * np.pi * np.pi) * sc.reciprocal(12 * lam * lam)\n",
+    "   \n",
+    "    return sc.DataArray(sc.sqrt(f * (a1 + a2 + a3)) + (q * std_dev_lam), coords={'wavelength':lam, 'spectrum':d.coords['spectrum'].copy()})\n",
```
SimonHeybrock: - Missing square on `q`.
- Parenthesis of `sqrt` end too early. But probably we should just return `sigma_Q^2` anyway, so remove the sqrt?

sans/reduction.ipynb
```diff
@@ -261,10 +302,13 @@
     "        dims=['Q'],\n",
     "        unit=sc.units.one/sc.units.angstrom,\n",
     "        values=np.geomspace(0.008, 0.6, num=55))\n",
-    "    # TODO QResolution\n",
+    "   \n",
     "    d = sc.Dataset({'data':data, 'norm':solid_angle(data)*direct_beam})\n",
+    "    wav_edges = d.coords['wavelength'].copy()\n",
     "    to_bin_centers(d, 'wavelength')\n",
     "    d = sc.neutron.convert(d, 'wavelength', 'Q', out=d) # TODO no gravity yet\n",
+    "    dq = q_resolution(wav_edges, moderator, d)\n",
```
SimonHeybrock: No need for copy of `wav_edges`, just move this up by two lines?

sans/reduction.ipynb
```diff
@@ -210,6 +210,55 @@
...
+    "    a2 = (r2*inv_l3)*(r2*inv_l3) * 3.0\n",
+    "    a3 = (dr/l2) * (dr/l2) \n",
+    "    q = 4.0 * np.pi * sc.sin(theta) * sc.reciprocal(lam) # keep in wav dim to prevent broadcast\n",
+    "    \n",
+    "    h = 6.62607004E-34\n",
+    "    m = 1.6749E-27\n",
+    "    k = sc.Variable(value=h/m, unit=(sc.units.m * sc.units.m)/sc.units.s) \n",
+    "    \n",
+    "    moderator.data.values *= 1e-6 # us to seconds\n",
+    "    moderator.data.unit = sc.units.s # fixup moderator units to microseconds. Previously dimensionless\n",
```
SimonHeybrock: Is this an bug in our Mantid converter?

sans/reduction.ipynb
```diff
@@ -210,6 +210,55 @@
...
+    "    \n",
+    "    # Calcualte moderator spread\n",
+    "    s = moderator.data * sc.reciprocal(l1 + l2) # times to inv velocity\n",
+    "    s *= k # inv velocity to wavelength\n",
+    "    s.unit = sc.units.angstrom\n",
+    "    s *= 1e10\n",
```
SimonHeybrock: Interesting! Looking in some more detail at the moderator spread calculation shows that this is simply a conversion from TOF to wavelength, except that the time is in the values of `moderator`, not in the coord. Please verify, but if true we can do something like

```python
mod = sc.Dataset(coords={'tof':moderator.data, 'source_position':moderator.attrs['source_position'], ...)
s = sc.neutron.convert(mod, 'tof', 'wavelength').coords['wavelength']
```
Avoiding the need for duplicating physical constants and unit conversions.

Interestingly, this hints at a wrong abstraction for unit conversions: Maybe unit conversion should be exposed in a different way, such that it cannot just convert coords. Will need to think about ths.

sans/reduction.ipynb
OwenArnold: Yes, removing sq then sqrt was an optimisation (same result), but probably didn't improve performance much. Returning sq would also be OK. Also given the discussion about how to apply to q1d, I might need to think about the boundaries of this function - as in is returning a mantid like result that might not be necessary.

sans/reduction.ipynb
OwenArnold: Yes. I had originally thought I could use the q calculated and set on d as part of unit conversion, but I cannot, so yes copy can now be avoided.

sans/reduction.ipynb
SimonHeybrock: It cannot be the same result, aren't the parenthesis in the wrong place?

sans/reduction.ipynb
OwenArnold: Indeed, would be nice if this was not limited to coordinates.

SimonHeybrock: @OwenArnold Try if this gives the same result:

```python
tof = moderator.data.copy()
tof.variances = None # shortcoming of Mantid or Mantid converter?
tof.rename_dims({'wavelength':'tof'}) # TODO overly restrictive check in convert (rename)
mod = sc.Dataset(coords={
    'tof':tof,
    'position':sample.coords['position'],
    'source_position':sample.coords['source_position'],
    'sample_position':sample.coords['sample_position']})
sc.neutron.convert(mod, 'tof', 'wavelength').coords['wavelength']
```

sans/reduction.ipynb
```diff
@@ -210,6 +210,55 @@
...
+    "    \n",
+    "    d_lam = lam_edges['wavelength', 1:] - lam_edges['wavelength', :-1] # bin widths\n",
+    "    lam = 0.5 * (lam_edges['wavelength', 1:] + lam_edges['wavelength', :-1]) # bin centres\n",
+    "    \n",
+    "    r2 = sc.Variable(value=4.0824829046386295/1000, unit=sc.units.m) # sample aperture radius\n",
+    "    r1 = sc.Variable(value=14.433756729740645/1000, unit=sc.units.m) # source aperture radius\n",
+    "    dr = sc.Variable(value=8.0/1000, unit=sc.units.m) # virtual ring width on detector\n",
+    "    \n",
+    "    l2 = sc.neutron.l2(d)\n",
+    "    l1 = sc.Variable(value=5.0, unit=sc.units.m) # should be sc.neutron.l1(d) = 25.6m ?\n",
```
SimonHeybrock: I think this is just a misleading naming in the equation, it should *not* be l1, but the collimation length, as specified for the algorithm. We *do* need the actual L1 for the moderator calculation though (but see the snippet I had suggested to avoid this altogether).

sans/reduction.ipynb
OwenArnold: I'll take a look at what the original outputs were. I think they might have been dimensionless in the `LoadRKH` generated workspace, but if not correct units could be attached in the converter.

sans/reduction.ipynb
OwenArnold: I'll change the name and remove my comment to avoid confusion.

sans/reduction.ipynb
```diff
@@ -210,6 +210,57 @@
...
+    "    \n",
+    "    d_lam = lam_edges['wavelength', 1:] - lam_edges['wavelength', :-1] # bin widths\n",
+    "    lam = 0.5 * (lam_edges['wavelength', 1:] + lam_edges['wavelength', :-1]) # bin centres\n",
+    "    \n",
+    "    r2 = sc.Variable(value=4.0824829046386295/1000, unit=sc.units.m) # sample aperture radius\n",
+    "    r1 = sc.Variable(value=14.433756729740645/1000, unit=sc.units.m) # source aperture radius\n",
+    "    dr = sc.Variable(value=8.0/1000, unit=sc.units.m) # virtual ring width on detector\n",
+    "    \n",
+    "    l2 = sc.neutron.l2(d)\n",
+    "    l_collimation = sc.Variable(value=5.0, unit=sc.units.m) # should be sc.neutron.l1(d) = 25.6m ?\n",
```
SimonHeybrock: Just make these parameters to `q_resolution`, as for the original algorithm?

SimonHeybrock: @OwenArnold I meant all the hard-coded values such as `r1`, etc..

PR #24, opened 2020-06-01
Adds radiography requirements 1. and 2.a

DavidFair:
This includes averaging out loaded data (for multiple reference images)
and transmission calculation.

The ROI code is WIP, however the transmission code causes us to run OOM,
so it would be better to figure out a way of resolving that first

One approach at the trade-off of space for speed is to calculate and apply the transmission correction projection by projection, rather than calculate it for the ~2000 projections in one

### DIFF
--- a/scipp-radiography.ipynb
+++ b/scipp-radiography.ipynb
@@ -1,0 +1,175 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "collapsed": true,
+    "pycharm": {
+     "name": "#%% md\n"
+    }
+   },
+   "source": [
+    "# How to start\n",
+    "\n",
+    "Before starting you must:\n",
+    "- Ensure that `scipp` and `mantid` are on your `PYTHONPATH`.\n",
+    "- Generate the `config.py` file using `make_config.py`. Refer to the `README.md` or `python make_config.py --help` for information.\n",
+    "\n",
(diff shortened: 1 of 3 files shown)


scipp-radiography.ipynb
```diff
@@ -0,0 +1,175 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "outputs": [],
+   "source": [
+    "dark_field = radiography.fits_to_variable(dark_field_dir, average_data=True)\n",
+    "flat_field = radiography.fits_to_variable(flat_field_dir, average_data=True)\n",
+    "\n",
+    "# This blows through 32GB of memory as raw_data[\"sample\"] is already ~17 GB\n",
```
OwenArnold, line 121: Can you implement below in place to avoid memory overhead of temporaries?
```
sample -= dark
flat -= dark
sample /= flat
```
or similar

scipp-radiography.ipynb
```diff
@@ -0,0 +1,175 @@
...
+    "pycharm": {
+     "name": "#%% md\n"
+    }
+   },
+   "source": [
+    "# How to start\n",
+    "\n",
+    "Before starting you must:\n",
+    "- Ensure that `scipp` and `mantid` are on your `PYTHONPATH`.\n",
+    "- Generate the `config.py` file using `make_config.py`. Refer to the `README.md` or `python make_config.py --help` for information.\n",
```
OwenArnold, line 16: Download data (specify own-cloud location) as a step?

PR #25, opened 2020-06-17
Direct beam iteration

SimonHeybrock:
Add first version of direct-beam iteration script. Includes the following:

- `loki.py`: Helper that knows the current geometry used for tests at LARMOR. This is likely going to be extended in the future.
- `sans.py`: Helper for SANS reduction. Not fully generic at this point, e.g., containing hard-coded knowledge about monitors. Should probably move those to attributes in the future.
- `contrib.py`: Helpers for things that are generic, but not possible with scipp yet. At the same time it was not clear if and how those should be added to scipp, so until further discussion and requirement gathering we keep this here.
- The actual notebook. This includes a demo of some useful visualizations, in particular for setting up masking. This part may later be extracted to a separate notebook (at least for the more static parts of the masks). Towards the end we plot various intermediate terms, which may be useful for debugging the iteration.
- The poly-gauss-coil fit is not my work but copied from the original workflow.

Note: This requires updated files. The ones from https://github.com/ess-dmsc-dram/loki_tube_scripts work, but the results look a bit weird. Will ask Judith to make the updated ones more genrally available.


sans/direct-beam.ipynb
```diff
@@ -0,0 +1,529 @@
...
+    "    return direct_beam"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import dataconfig # run make_config.py to create this\n",
```
OwenArnold: maybe catch import error and provide this information. Judith got stuck here.

sans/direct-beam.ipynb
```diff
@@ -0,0 +1,529 @@
...
+    "    print('start iteration')\n",
+    "    direct_beam_by_pixel = sc.choose(layers, choices=direct_beam, dim='layer')\n",
+    "    sample_q_lambda = to_q_by_wavelength(data=sample,\n",
+    "                                         transmission=sample_trans,\n",
+    "                                         direct_beam=direct_beam_by_pixel,\n",
+    "                                         direct_beam_transmission=background_trans, # note: background_trans\n",
+    "                                         masks=masks,\n",
+    "                                         wavelength_bands=wavelength_bands)\n",
+    "    background_q_lambda = to_q_by_wavelength(data=background,\n",
+    "                                             transmission=background_trans,\n",
```
OwenArnold, line 102: variables such as `background_trans` from global scope, should be passed as parameters.

sans/direct-beam.ipynb
```diff
@@ -0,0 +1,529 @@
...
+   "source": [
+    "pos = sc.neutron.position(sample)\n",
+    "x = sc.geometry.x(pos)\n",
+    "y = sc.geometry.y(pos)\n",
+    "counts = spectrum_counts.data\n",
+    "counts.variances = None # TODO support comparison with variances\n",
+    "masks['electronics-error'] = sc.less(sc.abs(x), 0.2 * sc.units.m) \\\n",
+    "                             & sc.less(sc.abs(y), 0.03 * sc.units.m) \\\n",
+    "                             & sc.equal(counts, 0.0 * sc.units.counts)\n",
+    "print(f\"Masking {sc.sum(masks['electronics-error'], 'spectrum').value} bad pixels due to electronics error.\")"
```
OwenArnold, line 292: Very nice logical form of combining and applying spacial and value based mask critera.

I believe this produces no masked elements in it's current form with the input data from the notebook. Why do we need it?

sans/direct-beam.ipynb
```diff
@@ -0,0 +1,529 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "scrolled": false
+   },
+   "outputs": [],
+   "source": [
+    "plot(pixel_counts, vmax=100, axes=['straw', 'tube', 'pixel'])"
```
OwenArnold, line 275: Slightly confusing that the plots here are not using the masks set up in previous steps. Maybe rearrange to move mask generation closer to point of use?

OwenArnold, line 275: Using dimensions in scipp like this looks excellent to me. I hope our users like the benefits of being able to do things like this.

sans/direct-beam.ipynb
```diff
@@ -0,0 +1,529 @@
...
+   "source": [
+    "masks['tube-ends'] = sc.greater(x, 0.36 * sc.units.m) \\\n",
+    "                   | sc.less(x, -0.36 * sc.units.m)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### X-Y projection\n",
```
OwenArnold, line 335: Not clear to me why this is required.

sans/direct-beam.ipynb
```diff
@@ -0,0 +1,529 @@
...
+    "sample.coords['low-counts'] = sc.less(counts, 100.0*sc.units.counts)\n",
+    "loki.instrument_view(sc.groupby(sample, 'low-counts').copy(group=1))\n",
+    "del sample.coords['low-counts']"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Reduction and direct-beam iteration"
```
OwenArnold, line 421: What's being done below here is sufficiently complex that it could use some explanation for me. Though that is probably not required for our intended audience of scientists.

OwenArnold (changes requested):
Some minor suggested changes. Will need to review direct-beam-iteration aspects.

sans/direct-beam.ipynb
SimonHeybrock: It does with the updated files (not on github).

sans/direct-beam.ipynb
SimonHeybrock: The main purpose of this was showing some tools that could be used during, e.g., commissioning. I am not adding the masks to the data array because actually we later need it for another data arrays as well (sample and background). I thought adding the masks to both would have distracted from the point.

sans/direct-beam.ipynb
SimonHeybrock: It is not (for the purpose of this script), just showing off. :)

It is "required" as a visualization of raw that that is more useful to instrument scientists and users: With 3D / volumetric detectors the instrument view has limited use for getting a quick preview of the data (it is probably more of a debugging / commissioning tool?). This is something that has been requested in the past.

sans/direct-beam.ipynb
SimonHeybrock: :+1:

SimonHeybrock:
Many more things to improve, but this is a working version which scientists are beginning to so, so I'll merge this.

PR #34, opened 2021-01-19
Fix loki.to_logical_dims after bugfix in scipp.reshape

SimonHeybrock:
Previously the reshape added a TOF extent of, e.g., 1000 to the coords. After a fix in scipp this bad behavior is not possible any more.

### DIFF
--- a/sans/loki.py
+++ b/sans/loki.py
@@ -49,33 +49,25 @@ class LoKI:
         Reshape data to use ['tube','straw','pixel'] instead of ['spectrum'].
         """
         import scipp as sc
-        dims = data.dims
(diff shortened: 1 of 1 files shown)


OwenArnold (commented):
Since we now using `ess-notebooks` and `ess` scipp org repositories in preference to `ess-legacy` and the fix be ported to those?

SimonHeybrock:
`scipp/ess/loki` does not appear to contain this component yet, this is why I changed it here. I think we didn't move it since it requires cleanup before inclusion in a published package.

OwenArnold (approved):

PR #38, opened 2021-08-27
New algorithms for diffraction

celinedurniak:
Python scripts: fft_smooth, remove_prompt_pulses, save_xye
Demo notebooks for calibration, fft_smooth and remove_prompt_pulses

### DIFF
--- a/diffraction/fft_smooth.py
+++ b/diffraction/fft_smooth.py
@@ -1,0 +1,70 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2021 Scipp contributors (https://github.com/scipp)
+# flake8: noqa: E501
+"""
+This script implements a few functions in order to perform a simple data reduction of POWGEN data
+using scipp.
+"""
+
+import numpy as np
+import scippneutron as scn
+from scipy.fft import irfft
+from scipy.fft import rfft
+
+
+def fft_smooth(dataset, n, order):
+    """
+    Performs smoothing of spectra using Butterworth filter 
+
+    dataset: input scipp histogrammed Dataset
+
+    n: integer to define cutoff
+
+    order: order of the Butterworth filter
+    """
+    
+    # n, order must be positive
+    assert isinstance(n, int), 'n should be an integer'
+    assert isinstance(order, int), 'order should be an integer'
+    assert (n > 0 and order > 0), 'n and order should be positive'
+
+    # coordinates of input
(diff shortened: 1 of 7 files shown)


diffraction/remove_prompt_pulses.py
```diff
@@ -0,0 +1,79 @@
...
+    frequency: frequency of the source (in Hz) used to calculate the minimum time 
+               of flight to filter.
+               If not provided it is calculated from the attribute in the dataset
+    """
+    # check inputs
+    assert width > 0, 'the width has to be positive'
+
+    # dataset must have tof as 1 of its coordinates
+    assert 'tof' in list(dataset.dims), 'tof must be one of the coordinates of dataset'
+    assert dataset.coords['tof'].unit == sc.units.us, 'tof must be expressed in microseconds'
```
jl-wynen, line 48: Do not require a specific unit, instead use `tof = sc.to_unit(da.coords['tof'], sc.units.us, copy=False)`.
Or, even better if possible, do not specify any units and let scipp figure everything out. This only works if you can avoid using numpy or raw numbers.

diffraction/remove_prompt_pulses.py
```diff
@@ -0,0 +1,79 @@
...
+    
+    if not frequency:
+        frequency = np.median(dataset.attrs['frequency'].values.values)
```
jl-wynen, line 51: Do not hard code the `'frequyency'` name but let the `frequency` function argument be either a `sc.Variable` or a `str`.

diffraction/remove_prompt_pulses.py
```diff
@@ -0,0 +1,79 @@
...
+
+    # period, min and max tof in microseconds
+    period = 1000000. / frequency 
+    tof_min = np.min(dataset.coords['tof'].values)  
+    tof_max = np.max(dataset.coords['tof'].values)
```
jl-wynen, line 56: Use `sc.min` and `sc.max`.

diffraction/fft_smooth.py
```diff
@@ -0,0 +1,70 @@
...
+using scipp.
+"""
+
+import numpy as np
+import scippneutron as scn
+from scipy.fft import irfft
+from scipy.fft import rfft
+
+
+def fft_smooth(dataset, n, order):
```
jl-wynen, line 15: Specify dimension to smooth in as an argument.
Then, there is no need to check for `'tof'` or `'spectrum'` below.

PR #39, opened 2022-01-14
Update scripts to scippneutron v0.4.1

celinedurniak:


diffraction/powder_reduction.py
```diff
@@ -186,10 +182,17 @@ def powder_reduction(sample='sample.nxs',
...
+        sample_dspacing = scn.convert(sample_lambda, 'wavelength', 'dspacing', scatter=True)
     else:
         # Calculate dspacing from calibration file
         sample_dspacing = convert_with_calibration(sample_lambda, cal_sample)
+
+        # remove wavelength coordinate and set d-spacing as one of the main ones
+        sample_dspacing = sample_dspacing.rename_dims({'wavelength': 'dspacing'})
+        del sample_dspacing.coords['wavelength']  # free up some memory, if not needed any more
+        if sample_dspacing.bins is not None:
+            del sample_dspacing.bins.coords['wavelength']  # free up a lot of memory
```
jl-wynen: You can use `transform_coords` with `keep_inputs=False` to remove wavelength more easily.
But this also removed the other inputs like `tzero`...

diffraction/powder_reduction.py
```diff
@@ -186,10 +182,17 @@ def powder_reduction(sample='sample.nxs',
     # 3. Convert to d-spacing with option of taking calibration into account
     if calibration is None:
         # No calibration data, use standard convert algorithm
-        sample_dspacing = scn.convert(sample_lambda, 'tof', 'dspacing')
+        sample_dspacing = scn.convert(sample_lambda, 'wavelength', 'dspacing', scatter=True)
```
jl-wynen: Why didn't you use `transform_coords` here?

jl-wynen (commented):
Looks ok. But note that scipp 0.11 introduces some more breaking changes.

Are you planning to move parts of this to `ess`?

diffraction/powder_reduction.py
celinedurniak: Using `scipp` v0.11.1 and `scippneutron` v0.4.2, I tried setting `keep_inputs` to `True` and `False`. The output datasets are identical except for the number of attributes (if `keep_inputs=True`, `difa`, `difc` are moved from coords to attributes and `tof` is kept as an attribute)
Screenshot 2022-01-19 at 08 18 38
.

diffraction/powder_reduction.py
celinedurniak: I removed the option to reduce data without calibration. Instead an exception is raised if equal to `None`.

Allowing `calibration=None` would require too many modifications because it is used for the calibration of detectors but also for the definition of groups, which is required to select the relevant groups of sample and vanadium data before normalisation. 

This option will be available in a more complete workflow.

celinedurniak:
The plan is to move some of these algorithms to the ess repo.

diffraction/powder_reduction.py
jl-wynen: Ah, right. You are using the full conversion graph. So it computes dspacing from tof and not from wavelength. Is this what you want?

If you want the latter, you need to either construct a graph yourself or use `scn.convert(target='dspacing', origin='wavelength')`. We don't expose a graph for that conversion yet.

celinedurniak: I want to convert to `dspacing` but with calibration. So I cannot use  `scn.convert(target='dspacing', origin='wavelength')`.

I can either 
convert from `wavelength` to `tof` (standard graph) and then convert from `tof` to `dspacing` using `calibration_graph`
or define a graph to convert from `wavelength` to `dspacing` with calibration

diffraction/powder_reduction.py
celinedurniak: I added an intermediate conversion to `tof` before applying `convert_with_calibration`.

diffraction/powder_reduction.py
celinedurniak: I added an intermediate conversion to `tof` before applying `convert_with_calibration`.

